PR #2, opened 2018-07-06
Make triehash generic over hasher

dvdplm:
Prepare `triehash` for use in Polkadot.

### DIFF
--- a/triehash/src/lib.rs
+++ b/triehash/src/lib.rs
@@ -28,8 +28,6 @@ extern crate rlp;
 use std::collections::BTreeMap;
 use std::cmp;
 use elastic_array::{ElasticArray4, ElasticArray8};
-// use ethereum_types::H256;
-use hash::keccak;
 use hashdb::Hasher;
 use rlp::RlpStream;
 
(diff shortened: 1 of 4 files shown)


triehash/src/lib.rs
```diff
@@ -114,31 +124,39 @@ pub fn trie_root<I, A, B>(input: I) -> H256
...
 		  A: AsRef<[u8]>,
 		  B: AsRef<[u8]>,
+		  H: Hasher,
+		  <H as hashdb::Hasher>::Out: cmp::Ord + rlp::Encodable,
 {
 	let gen_input: Vec<_> = input
 		// first put elements into btree to sort them and to remove duplicates
 		.into_iter()
-		.map(|(k, v)| (keccak(k), v))
+		.map(|(k, v)| (H::hash(&k.as_ref()), v))
```
twittner: Just `k.as_ref()` should be enough.

triehash/src/lib.rs
```diff
@@ -114,31 +124,39 @@ pub fn trie_root<I, A, B>(input: I) -> H256
...
 		.collect::<BTreeMap<_, _>>()
 		// then move them to a vector
 		.into_iter()
-		.map(|(k, v)| (as_nibbles(&k), v) )
+		.map(|(k, v)| (as_nibbles(&k.as_ref()), v) )
```
twittner: Just `k.as_ref()` should be enough.

PR #7, opened 2018-07-10
Upstream changes to kvdb-rocksdb

dvdplm:
ref. https://github.com/paritytech/parity/pull/8963

Upstream commit cda535: db: remove wal disabling / fast-and-loose option


andresilva:
I do not have admin powers to forcefully merge this 😛

PR #10, opened 2018-07-11
Add readmes for parity-common crates

### LINKED ISSUE #5: Suggestion: add short READMEs for all crates

dvdplm:
Closes #5

### DIFF
--- a/triehash/README.md
+++ b/triehash/README.md
@@ -1,0 +1,2 @@
+This crate provides utility functions to validate and initialize tries using flexible input.
+It is used extensively in `parity-ethereum` to validate blocks (mostly transactions and receipt roots).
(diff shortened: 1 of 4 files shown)


triehash/README.md
```diff
@@ -0,0 +1,2 @@
+This crate provides utility functions to validate and initialize tries using flexible input.
```
niklasad1: Nitpick, rephrase, `flexible input` (it is too ambiguous IMO)

What about generic input?

dvdplm: Yeah good pt. I think I'll just drop the whole last part.

niklasad1: Sounds good to me!

PR #12, opened 2018-07-13
Add fixed-hash to parity-common

dvdplm:
Both `susbtrate` and `ethereum-types` use `fixed-hash` to build their hash types so it's getting moved here.

### DIFF
--- a/fixed-hash/src/lib.rs
+++ b/fixed-hash/src/lib.rs
@@ -1,0 +1,40 @@
+// Copyright 2015-2017 Parity Technologies
+//
+// Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
+// http://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+// <LICENSE-MIT or http://opensource.org/licenses/MIT>, at your
+// option. This file may not be copied, modified, or distributed
+// except according to those terms.
+
+#![cfg_attr(not(feature = "std"), no_std)]
+
+#[cfg(all(feature="libc", not(target_os = "unknown")))]
+#[doc(hidden)]
+pub extern crate libc;
+
+#[cfg(not(feature="libc"))]
+#[doc(hidden)]
+pub mod libc { }
+
+#[cfg(feature="heapsizeof")]
+#[doc(hidden)]
+pub extern crate heapsize;
+
+#[cfg(feature="std")]
+#[doc(hidden)]
+pub extern crate core;
+
+#[cfg(feature="std")]
+#[doc(hidden)]
+pub extern crate rustc_hex;
+
+#[cfg(feature="std")]
+#[doc(hidden)]
+pub extern crate rand;
+
+#[cfg(feature="impl_quickcheck_arbitrary")]
+#[doc(hidden)]
+pub extern crate quickcheck;
+
+mod hash;
+pub use hash::*;
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,5 +1,6 @@
 [workspace]
 members = [
+	"fixed-hash",
 	"hashdb",
 	"keccak-hash",
 	"kvdb",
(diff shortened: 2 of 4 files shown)


tomusdrw:
I don't really get the reasoning for the move. The crate only exposes a macro to generate your own types, so it's only referenced by at most one crate.

`paritytech/primitives` already contains all the crates required to build uint/hashes, and doesn't depend on other git repos. Now `ethereum_types` for instance have to pull in `parity-common` repo?

None of the crates in this repo will ever dependend on `fixed-hash`, so I don't get what do we gain by having it here instead of in `primitives`.

IMHO we etiher keep `parity-common` and `primitives` separate or we dissolve `primitives` and move ALL crates here. Moving only `fixed-hash` doesn't make much sense to me.

fixed-hash/src/lib.rs
```diff
@@ -0,0 +1,40 @@
...
+#[doc(hidden)]
+pub mod libc { }
+
+#[cfg(feature="heapsizeof")]
+#[doc(hidden)]
+pub extern crate heapsize;
+
+#[cfg(feature="std")]
+#[doc(hidden)]
+pub extern crate core;
```
dvdplm: Why are all these crates `pub`?

uint/README.md
```diff
@@ -0,0 +1,8 @@
+# Big unsigned integer types
+
+Implementation of a various large-but-fixed sized unsigned integer types.
+The functions here are designed to be fast. There are optional `x86_64`
+implementations for even more speed, hidden behind the `x64_arithmetic`
+feature flag.
+
+Run tests with `cargo test --features=std,impl_quickcheck_arbitrary`.
```
dvdplm: How can I improve this? :/

twittner: Does https://doc.rust-lang.org/cargo/reference/manifest.html#the-required-features-field-optional help?

twittner:
Like @tomusdrw I am somewhat puzzled by this. Would the motivation behind `parity-common` not make it logical to put all `primitive` crates in here? But what prevents `parity-common` from becoming a black hole of crates? What about other repositories, e.g. `paritytech/exit-future`, `paritytech/bigint`, etc.? I guess the question I have is: When should a crate be a member of `parity-common` instead of being in a separate repository?

dvdplm:
> Would the motivation behind parity-common not make it logical to put all primitive crates in here?

Some of the code in `primitives` is ethereum-specific and should thus **not** be in `parity-common`.

> what prevents parity-common from becoming a black hole of crates? 

This is a great question and a concern I have too. The answer is, I guess, "we're disciplined about what we move". And "one could argue that `parity-ethereum` is already a black hole of crates and at least this new black hole is built with the strict requirement of code being used in more than one project".

The background conversation here is the tension between two valid but conflicting ergonomical concerns: 
1. splitting out code into own repos is great for consumers (easy to version, publish, and pull into other projects)
2. maintaining and updating a crate is much more comfortable when changes can be made in a single repo and with a single PR

The `parity-common` effort tries to find a middle ground between the two. And as mentioned, the criterium used is pretty pragmatic: "crates used in more than one Parity project that are not not already split out into their own repos".
As mentioned above I am also worried that we'll end up with big pile of unrelated crates and be back at square one, but I am also convicned that leaving things as they are is not a good option: we really really don't want to maintain forks and/or risk loosing track of what code is used where.

uint/README.md
dvdplm: I tried that and no, it's not really doing what I'd like. With:

```toml
[[test]]
name = "uint_tests"
required-features = ["std,impl_quickcheck_arbitrary"]
```
…I can run `cargo test` and all tests are skipped which is better than spewing errors, but now `cargo test --features=std,impl_quickcheck_arbitrary` also skips all tests.

What I'm looking for is a way to tell cargo "when running `cargo test` use this set of features: … … ".

dvdplm: **UPDATE**: seems to be a [known limitation](https://github.com/rust-lang/cargo/issues/2911)

dvdplm:
@tomusdrw @twittner @debris do you still have concerns about this?

PR #17, opened 2018-08-08
Remove HeapSizeOf-bound for Hasher::Out and add it only where needed

dvdplm:
The `HeapSizeOf` bound is causing trouble when [integrating `parity-common` with polkadot](https://github.com/paritytech/polkadot/pull/297) and is probably best added only where needed.


tomaka:
More generally, `heapsize` has been determined by their author to be a bad idea and should eventually be removed from all our code bases.

dvdplm:
@tomaka I for one is hating it with a vengeance. So, is that a 👍🏿 from you?

plain_hasher/util/plain_hasher/Cargo.toml
```diff
@@ -1,13 +0,0 @@
...
-version = "0.1.0"
-authors = ["Parity Technologies <admin@parity.io>"]
-license = "MIT"
-keywords = ["hash", "hasher"]
-homepage = "https://github.com/paritytech/plain_hasher"
-
-[dependencies]
-crunchy = "0.1.6"
-ethereum-types = "0.3"
-hashdb = { version = "0.2.0", path = "../hashdb" }
```
debris: what happened with this file? ;)

dvdplm: It was a left-over from when I moved over the code from `parity-ethereum` – I must have messed up the `git subtree` command at some point.

PR #18, opened 2018-08-08
unbound salt size

tworec:
There is no reason to enforce this 32 bytes bound. Salt can have arbitrary lenght.


dvdplm:
Did you check if consuming code (e.g. https://github.com/paritytech/parity-ethereum/blob/master/ethstore/src/account/crypto.rs#L87) works with an unbounded salt? In other words, is this a backwards incompatible change?

Also, can you elaborate on your use-case driving this change? What problem did you have that needed this change? 

I get that salt-size isn't particularly security sensitive, but this change allows it to be arbitrarily short (as well as long) and 1-byte long salts could be a problem.

PR #22, opened 2018-08-09
[rlp] make it possible to use rlp without pulling in ethereum-types

dvdplm:
Makes it possible to use `rlp` in polkadot without pulling in `ethereum-types` and associated luggage.

### DIFF
--- a/rlp/src/impls.rs
+++ b/rlp/src/impls.rs
@@ -8,6 +8,7 @@
 
 use std::{cmp, mem, str};
 use byteorder::{ByteOrder, BigEndian};
+#[cfg(feature = "ethereum")]
 use bigint::{U128, U256, H64, H128, H160, H256, H512, H520, Bloom};
 use traits::{Encodable, Decodable};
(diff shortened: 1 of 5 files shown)


rlp/src/impls.rs
```diff
@@ -207,20 +208,34 @@ macro_rules! impl_decodable_for_hash {
 	}
 }
 
+#[cfg(feature = "ethereum")]
 impl_encodable_for_hash!(H64);
```
debris: can we move all of these macro calls into a separate module to avoid writing `cfg` 20 times? :)

rlp/src/lib.rs
```diff
@@ -33,9 +33,13 @@
 //! * You don't want to decode whole rlp at once.
 
 extern crate byteorder;
+#[cfg(feature = "ethereum")]
 extern crate ethereum_types as bigint;
```
debris: bigint? lol. It's probably done by me when refactoring and I forgot to remove the alias

PR #25, opened 2018-08-10
Port over relevant changes from bigint

dvdplm:
The `bigint` crate is going away but there is valuable code there that is ported over in this PR.

During the recent extraction of crates from `parity-ethereum` and `primitives`, the `uint` crate does not instantiate any types, leaving that for consuming code to take care of (e.g. `ethereum-types` or `substrate/primitives`). This means that some code in `bigint` cannot be ported to `uint` but should rather be checked against what `ethereum-types` has today. Examples of this is the `impl U256 {…}`, the serde serialization stuff and the many conversion impls in `bigint`.

The `bigint` crate adds a [`mod_inverse`](https://github.com/paritytech/bigint/blob/master/src/uint.rs#L988) function to `U256`. I think this is the only feature not present in `uint`. If it turns out we need this, it should probably be added to `ethereum-types`.

The repos have diverged significantly so this port has been made manually. I am sure I have missed things, so thorough review by the `bigint` committers would be very useful.

**Performance**

The `uint` crate did not have any benchmarks so this PR copies over the ones from `bigint`. The results are odd: `uint` is much faster across the board. This was unexpected and makes me suspect I did something wrong.

Results for `uint`:

```
running 12 tests
test u128_mul       ... bench:      10,466 ns/iter (+/- 649)
test u256_add       ... bench:       5,205 ns/iter (+/- 313)
test u256_from_be   ... bench:           0 ns/iter (+/- 0)
test u256_from_le   ... bench:           4 ns/iter (+/- 0)
test u256_full_mul  ... bench:      39,983 ns/iter (+/- 2,418)
test u256_mul       ... bench:      20,673 ns/iter (+/- 1,583)
test u256_mul_small ... bench:         144 ns/iter (+/- 7)
test u256_sub       ... bench:       5,248 ns/iter (+/- 370)
test u512_add       ... bench:      22,656 ns/iter (+/- 1,158)
test u512_mul       ... bench:      49,118 ns/iter (+/- 2,795)
test u512_mul_small ... bench:         587 ns/iter (+/- 37)
test u512_sub       ... bench:      24,038 ns/iter (+/- 1,502)

test result: ok. 0 passed; 0 failed; 0 ignored; 12 measured; 0 filtered out
```

Results for `bigint`:

```
running 12 tests
test u128_mul       ... bench:      15,753 ns/iter (+/- 921)
test u256_add       ... bench:      21,017 ns/iter (+/- 1,666)
test u256_from_be   ... bench:           4 ns/iter (+/- 0)
test u256_from_le   ... bench:           6 ns/iter (+/- 0)
test u256_full_mul  ... bench:      40,166 ns/iter (+/- 2,646)
test u256_mul       ... bench:      67,205 ns/iter (+/- 4,982)
test u256_mul_small ... bench:         489 ns/iter (+/- 47)
test u256_sub       ... bench:      26,734 ns/iter (+/- 3,208)
test u512_add       ... bench:      22,578 ns/iter (+/- 1,352)
test u512_mul       ... bench:     320,132 ns/iter (+/- 18,449)
test u512_mul_small ... bench:       3,700 ns/iter (+/- 314)
test u512_sub       ... bench:      23,576 ns/iter (+/- 1,819)
```


uint/src/uint.rs
```diff
@@ -1219,16 +1195,24 @@ macro_rules! impl_std_for_uint {
 					return write!(f, "0");
 				}
 
-				let mut s = String::new();
+				let mut buf = [0_u8; $n_words*20];
```
Vurich: Why `$n_words * 20`? Surely it's `ceil(log10(2^($n_words - 1)))`? That's 154 for U512 but this code would create a buffer of size 1280 (way too big).

uint/src/uint.rs
```diff
@@ -1281,14 +1264,25 @@ macro_rules! impl_std_for_uint {
...
+	);
 }
 
-#[cfg(not(feature="std"))]
+#[cfg(feature="std")]
 #[macro_export]
 #[doc(hidden)]
-macro_rules! impl_std_for_uint {
-	($name: ident, $n_words: tt) => {}
+macro_rules! impl_std_for_uint_internals {
```
Vurich: Why change the name of this macro? If it's used by another macro then you should use a private pattern of that macro rather than a `doc(hidden)` macro.

uint/src/uint.rs
dvdplm: This change came from https://github.com/paritytech/bigint/blob/master/src/uint.rs#L927 (part of PR [#43](https://github.com/paritytech/bigint/pull/43)) – not sure what their thinking was there, maybe it was making space for a somewhat long debug string (which doesn't make sense).
Should I just go ahead and replace it?

Vurich: Yes please!

dvdplm: Wait a sec, I think you're confusing `uint` with `fixed-hash`: here the `construct_uint!` macro is called with the **words** not the number of bytes, so `$n_words` here is `8` => `8 * 10 == 160` which is fine I think.

dvdplm: The `construct_hash!` macro takes number of bytes, which is more logical imo. Quite confusing.

NikVolf:
why do we need the same code in 2 places anyway?

dvdplm:
> why do we need the same code in 2 places anyway?

Not sure I follow, can you elaborate? The idea here is to:

1. bring useful changes from `bigint` into `uint` and then 
2. stop using `bigint`

uint/Cargo.toml
```diff
@@ -4,27 +4,23 @@ homepage = "http://parity.io"
 repository = "https://github.com/paritytech/primitives"
```
NikVolf: btw, repository needs an update?

NikVolf:
You import code from MIT/Apache repository, btw

under GPL license while external contributors agreed under MIT/Apache contribution

Pzixel:
Hey guys. I contributed a bit but I don't mind of you copy my code under diffirent license.

About my changes in `no_std` format: `$n_words*20` is the smallest buffer you could use to format any uint of size `$n_words` losslessly. It's not just a round number or something. I don't recall exactly why you cannot use `$n_words*16` (which naively seems to be the correct smallest buffer size), but you can't.

dvdplm:
@Pzixel your code is correct, we were mistaken. :) And thank you for your contribution and openness to sort out the licensing stuff. Much appreciated!

Pzixel:
Great work!

I'd also suggest to review this discussion about implementing panicing `From` trait: https://github.com/paritytech/bigint/pull/43#discussion_r204255792

In a nutshell: rust documentation clearly says that `From` should not fail (and panic is obviosly a fail), when guys think it's ok becase you have a `'static` constraint on it. You probably want t revisit with decision before merging into master.

Vurich:
@Pzixel According to this simple Rust program:

```rust
fn main() {
    for &i in &[128, 256, 512] {
        println!("{}, {}", (i / 32) * 10, 2f64.powi(i - 1).log10().ceil());
    }
}
```

You could get away with doing `$n_words * 10` and it'd still work.

Vurich:
I think that the best way to handle the bad `From` impl is to either do `U256::try_from("123123").unwrap()` or have an inherent fn that takes a `&str` of any lifetime and documents that it panicks.

Vurich:
What error? It seems like you should get an OOB index and not a UTF8 error,

Pzixel:
@Vurich yep, this is why I edited my post. But you probably got an old version in email notification. Nevermind. Just check it yorself if you wish, it will grant better understanding and I will be calm that I didn't lie :)

dvdplm:
Using `$n_words * 10` does indeed fail.

Pzixel:
> I think that the best way to handle the bad From impl is to either do U256::try_from("123123").unwrap() or have an inherent fn that takes a &str of any lifetime with a doc comment saying that it may panic.

there is already `Parse` so you don't need any `try_from`, just `parse` :) See https://github.com/paritytech/bigint/pull/43/commits/918397684a2db21990a9e69571129b806aa5f911#diff-eb8a6a0b9d6d83979e35e6b02a6eafbfL2000

uint/benches/bigint.rs
```diff
@@ -0,0 +1,230 @@
...
+//! ```
+
+#![feature(test)]
+
+extern crate core;
+extern crate test;
+#[macro_use]
+extern crate crunchy;
+#[macro_use]
+extern crate uint;
```
andresilva: Add newline before macro invocation?

uint/benches/bigint.rs
```diff
@@ -0,0 +1,230 @@
...
+construct_uint!(U128, 2);
+construct_uint!(U256, 4);
```
andresilva: Can we use the exported types here? Apparently, creating the types here influences the benchmark results.

uint/benches/bigint.rs
andresilva: Add newline before macro invocation?

PR #28, opened 2018-08-20
Relax version requirements

dvdplm:
Omitting the "internal" version requirements for crates within `parity-common` should let downstream consumers set their own requirements. Also use less features where possible.

### DIFF
--- a/memorydb/Cargo.toml
+++ b/memorydb/Cargo.toml
@@ -8,9 +8,9 @@ license = "GPL-3.0"
 [dependencies]
 elastic-array = "0.10"
 heapsize = "0.4"
-hashdb = { version = "0.2.0", path = "../hashdb" }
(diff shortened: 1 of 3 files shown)


pepyakin:
afaik, crates.io doesn't accept `Cargo.toml`s with local deps without explicitly set version, so this change will prevent us to publish these crates as-is. Is that ok?

dvdplm:
> afaik, crates.io doesn't accept Cargo.tomls with local deps without explicitly set version, so this change will prevent us to publish these crates as-is. Is that ok?

Ouch, no that's not ok. :(

Reason I did this was I noticed consuming code (in polkadot) did not pick up the right version, but must have been for a different reason then. I'll revert those changes. Thanks!

NikVolf:
needs rebase

PR #30, opened 2018-08-21
Add conversion from [u8; *]

dvdplm:
Conversions I skipped when porting over stuff from bigint but that turned out to be useful.

### DIFF
--- a/uint/src/uint.rs
+++ b/uint/src/uint.rs
@@ -813,6 +813,19 @@ macro_rules! construct_uint {
 				arr
 			}
 		}
+
+		impl From<[u8; $n_words * 8]> for $name {
+			fn from(bytes: [u8; $n_words * 8]) -> Self {
+				bytes[..].as_ref().into()
+			}
+		}
+
+		impl<'a> From<&'a [u8; $n_words * 8]> for $name {
+			fn from(bytes: &[u8; $n_words * 8]) -> Self {
+				bytes[..].into()
+			}
+		}
+
 		impl Default for $name {
 			fn default() -> Self {
 				$name::zero()
--- a/uint/Cargo.toml
+++ b/uint/Cargo.toml
@@ -4,7 +4,7 @@ homepage = "http://parity.io"
 repository = "https://github.com/paritytech/parity-common"
 license = "MIT/Apache-2.0"
 name = "uint"
-version = "0.3.0"
+version = "0.3.1"
 authors = ["Parity Technologies <admin@parity.io>"]
 
 [dependencies]
--- a/uint/tests/uint_tests.rs
+++ b/uint/tests/uint_tests.rs
@@ -1001,7 +1001,7 @@ fn from_big_endian() {
 }
 
 #[test]
-fn from_fixed_array() {
+fn into_fixed_array() {
 	let expected: [u8; 32] = [
 		0, 0, 0, 0, 0, 0, 0, 0,
 		0, 0, 0, 0, 0, 0, 0, 0,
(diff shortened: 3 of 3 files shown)


uint/src/uint.rs
```diff
@@ -813,6 +813,19 @@ macro_rules! construct_uint {
 				arr
 			}
 		}
+
+		impl From<[u8; $n_words * 8]> for $name {
+			fn from(bytes: [u8; $n_words * 8]) -> Self {
+				bytes[..].as_ref().into()
```
fckt: just wonder.. Is it possible to convert without copy here?

dvdplm: You mean like this?

```rust
		impl From<[u8; $n_words * 8]> for $name {
			fn from(bytes: [u8; $n_words * 8]) -> Self {
				let bytes : [u64; $n_words] = unsafe { ::core::mem::transmute(bytes) };
				$name(bytes)
			}
		}
```

Benchmarking that is no speedup at all though. :/ 

Where is the copy you mention, I fail to see it?

Here's the benchmark I used, maybe it's wrong?

```rust
#[bench]
fn u512_from_fixed_array(b: &mut Bencher) {
    let ary512 : [u8; 64] =  [
        255, 0, 0, 123, 0, 0, 0, 0,
        0, 0, 0, 0, 0, 0, 0, 0,
        0, 0, 0, 121, 0, 0, 0, 0,
        0, 213, 0, 0, 0, 0, 0, 0,
        0, 0, 0, 0, 0, 0, 100, 0,
        0, 0, 0, 0, 0, 0, 0, 0,
        0, 45, 0, 0, 67, 0, 0, 0,
        0, 0, 0, 0, 0, 0, 0, 123
    ];
    let ary256 : [u8; 32] =  [
        255, 0, 0, 123, 0, 0, 0, 0,
        0, 0, 0, 0, 0, 0, 0, 0,
        0, 0, 0, 121, 0, 0, 0, 0,
        0, 213, 0, 0, 0, 0, 0, 0,
    ];
    b.iter(|| {
        let n = black_box(1000);
        for _i in 0..n {
            let _ : U512 = black_box(ary512).into();
            let _ : U256 = black_box(ary256).into();
        }
    })
}
```

fckt: as I understand, in should resolved to this call this: https://github.com/paritytech/parity-common/blob/6d47b026897bc940f97b1cb8ea56d79c3facda2a/uint/src/uint.rs#L777
so..

dvdplm:
I actually broke the impl in https://github.com/paritytech/parity-common/pull/30/commits/72e84868e800db99bb186e17d7dc4ef8783d4635 but https://github.com/paritytech/parity-common/pull/30/commits/283b13f11440bc861b7374ebbd6fcccc3b8ecf74 should fix it.

uint/src/uint.rs
```diff
@@ -813,6 +813,16 @@ macro_rules! construct_uint {
 				arr
 			}
 		}
+
+		impl From<[u8; $n_words * 8]> for $name {
+			fn from(bytes: [u8; $n_words * 8]) -> Self {
+				let mut bytes = bytes;
+				bytes[..].reverse();
```
fckt: but why we should reverse bytes? I guess, user expects that this conversion will just reinterpret  the bytes as `uint` without any additional transformations, am I wrong..? Or the reason behind it is that it just conflicts with former implementation?..

dvdplm: > the reason behind it is that it just conflicts with former implementation?.

Yeah. I have to be honest: I expected the first impl to be correct, but the tests failed miserably so I banged on it in a bunch of different ways until they passed... :/ So I *think* the reason is this: the U* types are defined as `[u64; $n_words]` and index 0 is the lower end of the range (see sample below) *but* within the `u64` the number is little endian (on x86), so the last byte is the smallest.

The problem with my explanation is that if I'm right, then it should be enough to reverse the array *after* transmuting it. So I must be missing something! :(

```rust
		impl From<u64> for $name {
			fn from(value: u64) -> $name {
				let mut ret = [0; $n_words];
				ret[0] = value;
				$name(ret)
			}
		}
```

NikVolf: U256 itself is little-endian

but default implementation from slice interprets _slice_ as big endian integer

i think you should use byteorder crate instead of unsafe transmute

uint/src/uint.rs
```diff
@@ -813,6 +813,16 @@ macro_rules! construct_uint {
 				arr
 			}
 		}
+
+		impl From<[u8; $n_words * 8]> for $name {
+			fn from(bytes: [u8; $n_words * 8]) -> Self {
```
NikVolf: why not just use https://github.com/paritytech/parity-common/blob/283b13f11440bc861b7374ebbd6fcccc3b8ecf74/uint/src/uint.rs#L777

?

NikVolf:
@niklasad1 

No, this should support any endianness

uint/src/uint.rs
fckt: @NikVolf https://github.com/paritytech/parity-common/blob/6d47b026897bc940f97b1cb8ea56d79c3facda2a/uint/src/uint.rs#L777

uint/src/uint.rs
dvdplm: Yes, I'm reverting this to just to `bytes[..].as_ref().into()` (which will resolve to using `from_big_endian`).

PR #32, opened 2018-08-22
Fix docs, remove outdated comments, remove overflowing divs/rems.

NikVolf:
(`overflowing_div`/`overflowing_rem` is pointless for unsigned integers)


sorpaas:
Not yet sure whether this is a good idea or not, but I just want to note that Rust's primitive type like `u64` do have `overflowing_div` and `overflowing_rem`. Would we want to consider to add them still as helper functions, just so that, quoting Rust's docs, "all operations are accounted for in the overflowing operations"?

PR #34, opened 2018-08-22
Port over relevant tests and code from parity-hash

dvdplm:
The `pwasm-*` crates currently use [`parity-hash`](https://github.com/fckt/parity-hash) while `parity-ethereum` and `substrate` use `fixed-hash`. This PR tries to rectify this by porting over relevant changes from `parity-hash` to `fixed-hash` with the end goal of deprecating the former. One important difference is that `fixed-hash` does not export any concrete types so in order to stop using `parity-hash` the type construction should be added to one of the `pwasm-*` crates (`pwasm-abi` perhaps?).

### DIFF
--- a/fixed-hash/src/hash.rs
+++ b/fixed-hash/src/hash.rs
@@ -344,8 +344,7 @@ macro_rules! impl_hash_conversions {
 	($a: ident, $a_size: expr, $b: ident, $b_size: expr) => {
 		impl From<$b> for $a {
 			fn from(value: $b) -> $a {
-				// REVIEW: better way of ensuring the macro params are ok?
-				assert!($a_size > $b_size && $a_size % 2 == 0 && $b_size %2 == 0);
+				debug_assert!($a_size > $b_size && $a_size % 2 == 0 && $b_size %2 == 0);
 				let mut ret = $a::new();
 				ret.0[($a_size - $b_size)..$a_size].copy_from_slice(&value);
 				ret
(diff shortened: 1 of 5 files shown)


fixed-hash/src/hash.rs
```diff
@@ -321,6 +335,80 @@ macro_rules! construct_hash {
...
+				let mut ret = $a::new();
+				ret.0[($a_size - $b_size)..$a_size].copy_from_slice(value);
+				ret
+			}
+		}
+	}
+}
+
+/// Implements conversion to and from a hash type and the equally sized unsigned int.
+/// CAUTION: Bad things will happen if the two types are not of the same size!
```
NikVolf: maybe add a debug_assert along with caution?

dvdplm: Sort of like I did [here](https://github.com/paritytech/rust-rocksdb)? What are your thoughts on `assert` vs `debug_assert` in this case?

NikVolf: regular `assert` is not zero cost in production i believe

dvdplm: Right, I was more asking if it's worth it to use `assert!` anyway or if the cost is prohibitive.

pepyakin: what bad things ought to happen exactly? Isn’t this bad things already encompass panicking?

dvdplm: @pepyakin you are correct! :) Added a `debug_assert`, which is similar to what `uint` does.

PR #37, opened 2018-08-23
rlp optimization

debris:
benches before:

```
test bench_decode_1000_values        ... bench:      46,870 ns/iter (+/- 4,621)
test bench_decode_nested_empty_lists ... bench:         381 ns/iter (+/- 59)
test bench_decode_u256_value         ... bench:          54 ns/iter (+/- 9)
test bench_decode_u64_value          ... bench:          71 ns/iter (+/- 11)
```

benches after:

```
test bench_decode_1000_values        ... bench:      39,273 ns/iter (+/- 3,948)
test bench_decode_nested_empty_lists ... bench:         381 ns/iter (+/- 49)
test bench_decode_u256_value         ... bench:          50 ns/iter (+/- 9)
test bench_decode_u64_value          ... bench:          58 ns/iter (+/- 6)
```

### DIFF
--- a/rlp/src/rlpin.rs
+++ b/rlp/src/rlpin.rs
@@ -72,21 +72,19 @@ impl PayloadInfo {
 
 	/// Create a new object from the given bytes RLP. The bytes
 	pub fn from(header_bytes: &[u8]) -> Result<PayloadInfo, DecoderError> {
-		match header_bytes.first().cloned() {
-			None => Err(DecoderError::RlpIsTooShort),
-			Some(0...0x7f) => Ok(PayloadInfo::new(0, 1)),
-			Some(l @ 0x80...0xb7) => Ok(PayloadInfo::new(1, l as usize - 0x80)),
-			Some(l @ 0xb8...0xbf) => {
-				let len_of_len = l as usize - 0xb7;
-				calculate_payload_info(header_bytes, len_of_len)
-			}
-			Some(l @ 0xc0...0xf7) => Ok(PayloadInfo::new(1, l as usize - 0xc0)),
-			Some(l @ 0xf8...0xff) => {
-				let len_of_len = l as usize - 0xf7;
-				calculate_payload_info(header_bytes, len_of_len)
-			},
-			// we cant reach this place, but rust requires _ to be implemented
-			_ => { unreachable!(); }
+		let l = *header_bytes.first().ok_or_else(|| DecoderError::RlpIsTooShort)?;
+		if l <= 0x7f {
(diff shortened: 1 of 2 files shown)


rlp/src/rlpin.rs
```diff
@@ -261,7 +249,7 @@ impl<'a, 'view> Rlp<'a> where 'a: 'view {
 	}
 
 	pub fn decoder(&self) -> BasicDecoder {
-		BasicDecoder::new(self.clone())
+		BasicDecoder::new(self.bytes)
```
debris: 1st optimization:
- do not copy `Cell`'s

rlp/src/rlpin.rs
```diff
@@ -342,43 +330,38 @@ impl<'a> BasicDecoder<'a> {
...
-				let len = decode_usize(&bytes[1..begin_of_value])?;
+		let bytes = self.rlp;
 
-				let last_index_of_value = begin_of_value.checked_add(len)
-					.ok_or(DecoderError::RlpInvalidLength)?;
-				if bytes.len() < last_index_of_value {
-					return Err(DecoderError::RlpInconsistentLengthAndData);
-				}
-				Ok(f(&bytes[begin_of_value..last_index_of_value])?)
+		let l = *bytes.first().ok_or_else(|| DecoderError::RlpIsTooShort)?;
```
debris: 2nd optimization:

- do not call `cloned()`

rlp/src/rlpin.rs
```diff
@@ -342,43 +330,38 @@ impl<'a> BasicDecoder<'a> {
...
+
+		if l <= 0x7f {
```
debris: 3rd optimization:

- avoid complex pattern matching

rlp/src/rlpin.rs
niklasad1: ```cloned() == into_iter().map(|&x| x.clone)``` ?

hrm, have you checked the assembly from `cloned()`?  I would expect this to be optimized to just a single indirection since the address to `bytes` should be in a register but I'm not sure!

`deref` is better here but shouldn´t impact ```performance``` (should produce the same asm), however it is more clear because `clone` is considered an expensive operation

niklasad1: also isn't more efficient with `ok_or` here because `DecoderError::RlpIsTooShort` is just a value? 

I don't know how expensive it is to create the closure :smile_cat:

debris: I haven't looked at asm. it was just ~3% faster every time I run `bench_decode_1000_values` bench

PR #40, opened 2018-08-24
Refactor OwnedNode::Branch

tafia:
This is a (small) continuation of #29.

Refactors the `OwnedNode::Branch` variant with a dedicated `Branch` struct.
Uses a single `Vec` to hold all the data. We save several `NodeKey`s and a `DBValue` which are all `ElasticArray`s.

This struct is immutable, which is ok I think reading at the comment for `OwnedNode` (mainly used for iterations). I do not grasp all the consequences out of parity-common.

As for the benches, as expected, iterating the trie is faster. I guess the memory should also be lower:
```
$ cargo benchcmp master branch
 name                        master ns/iter  branch ns/iter  diff ns/iter   diff %  speedup 
 trie_insertions_32_mir_1k   3,557,064       3,606,762             49,698    1.40%   x 0.99 
 trie_insertions_32_ran_1k   3,512,891       3,617,559            104,668    2.98%   x 0.97 
 trie_insertions_random_mid  2,618,781       2,605,743            -13,038   -0.50%   x 1.01 
 trie_insertions_six_high    2,688,483       2,675,930            -12,553   -0.47%   x 1.00 
 trie_insertions_six_low     5,206,615       5,177,413            -29,202   -0.56%   x 1.01 
 trie_insertions_six_mid     3,386,309       3,383,847             -2,462   -0.07%   x 1.00 
 trie_iter                   2,610,651       2,222,271           -388,380  -14.88%   x 1.17
```

### DIFF
--- a/patricia_trie/src/node.rs
+++ b/patricia_trie/src/node.rs
@@ -35,6 +35,33 @@ pub enum Node<'a> {
 	Branch([&'a [u8]; 16], Option<&'a [u8]>),
 }
 
+/// A Sparse (non mutable) owned vector struct to hold branch keys
+#[derive(Eq, PartialEq, Debug, Clone)]
+pub struct BranchKeys {
+	data: Vec<u8>,
+	ubounds: [usize; 17],
+}
+
+impl<'a> From<[&'a [u8]; 16]> for BranchKeys {
+	fn from(a: [&'a [u8]; 16]) -> Self {
+		let mut data = Vec::with_capacity(a.iter().map(|inner| inner.len()).sum());
+		let mut ubounds = [0; 17];
+		for (inner, ub) in a.iter().zip(ubounds.iter_mut().skip(1)) {
+			data.extend_from_slice(inner);
+			*ub = data.len();
+		}
+		BranchKeys { data, ubounds }
+	}
+}
+
+impl ::std::ops::Index<usize> for BranchKeys {
+	type Output = [u8];
+	fn index(&self, index: usize) -> &[u8] {
+		assert!(index < 16);
+		&self.data[self.ubounds[index]..self.ubounds[index + 1]]
+	}
+}
+
 /// An owning node type. Useful for trie iterators.
 #[derive(Debug, PartialEq, Eq)]
 pub enum OwnedNode {
@@ -45,7 +72,7 @@ pub enum OwnedNode {
 	/// Extension node: partial key and child node.
 	Extension(NibbleVec, DBValue),
 	/// Branch node: 16 children and an optional value.
-	Branch([NodeKey; 16], Option<DBValue>),
+	Branch(BranchKeys, Option<DBValue>),
 }
 
 impl<'a> From<Node<'a>> for OwnedNode {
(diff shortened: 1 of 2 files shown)


tafia:
Not sure where to ask so here I am.

How big are the `DBValue`, `NodeKeys` and other nibbles expected to be in practice?
How frequent is the DB modification vs iteration?

This is to get an idea and stop trying to change thing that are probably not very relevant (and make you lose your time).

patricia_trie/src/triedb.rs
```diff
@@ -365,27 +365,30 @@ impl<'a, H: Hasher, C: NodeCodec<H>> Iterator for TrieDBIterator<'a, H, C> {
...
 							},
-							OwnedNode::Branch(_, _) => { self.key_nibbles.pop(); },
+							OwnedNode::Branch(_) => { self.key_nibbles.pop(); },
 							_ => {}
 						}
 						IterStep::PopTrail
 					},
-					(Status::At, &OwnedNode::Leaf(_, ref v)) | (Status::At, &OwnedNode::Branch(_, Some(ref v))) => {
+					(Status::At, &OwnedNode::Branch(ref branch)) if branch.has_value() => {
+						return Some(Ok((self.key(), DBValue::from_slice(branch.get_value().unwrap()))));
```
dvdplm: Instead of scary-looking `unwrap()` can you use `expect("with the proof that this is ok, e.g. 'already checked that it has_value()'; qed")`?

patricia_trie/src/node.rs
```diff
@@ -35,6 +35,50 @@ pub enum Node<'a> {
 	Branch([&'a [u8]; 16], Option<&'a [u8]>),
 }
 
+/// A Sparse (non mutable) owned vector struct to hold branch keys and value
+#[derive(Eq, PartialEq, Debug, Clone)]
+pub struct Branch {
```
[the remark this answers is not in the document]
tafia: Yes because it is in a variant of another public struct

patricia_trie/src/triedb.rs
tafia: sure

patricia_trie/src/node.rs
```diff
@@ -45,7 +91,7 @@ pub enum OwnedNode {
 	/// Extension node: partial key and child node.
 	Extension(NibbleVec, DBValue),
 	/// Branch node: 16 children and an optional value.
-	Branch([NodeKey; 16], Option<DBValue>),
```
debris: is `NodeKey` used anywhere else?

tafia: > did you measure what was the impact of the performance if you only replaced ElasticArray with Vec? I'm just curious

No I didn't. There is neither `extend` nor `with_capacity` method for `ElasticArray`. I expect it to have bad performance, except if we build the vec on our side and use `from_vec` but it defeats the test.

In practice as I've said, I don't have a clear vision of what is the real situation, how big are all these collections? Do the bench make a good job at showing real scenarios?

tafia:
> did you measure what was the impact of the performance if you only replaced ElasticArray with Vec? I'm just curious

No I didn't. There is neither `extend` nor `with_capacity` method for `ElasticArray`. I expect it to have bad performance, except if we build the vec on our side and use `from_vec` but it defeats the test.

In practice as I've said, I don't have a clear vision of what is the real situation, how big are all these collections? Do the bench make a good job at showing real scenarios?

5chdn:
is this good to go?

tafia:
I believe so.

Just for information, because I don't know which channel is the best to discuss this, I am currently pushing the idea of allocating all data in a unique Vec for the whole trie [here](https://github.com/tafia/quick-patricia-trie). Don't trust benchmarks too much for now (in particular the iterator one, I suspect I am doing much less things).

PR #42, opened 2018-08-24
Fix scoring to favour lower insertion_ids.

tomusdrw:
Related to: https://github.com/paritytech/parity-ethereum/issues/9398

The issue was that current `ScoringWithRef` implementation was falling back to compare `insertion_id` in case the score is equal, but recently imported transactions were favored - this was incorrect, fifo behaviour is more fair and expected.

CC @Tbaut @dvdplm

### DIFF
--- a/transaction-pool/src/scoring.rs
+++ b/transaction-pool/src/scoring.rs
@@ -138,7 +138,7 @@ impl<T, S: Clone> Clone for ScoreWithRef<T, S> {
 impl<S: cmp::Ord, T> Ord for ScoreWithRef<T, S> {
 	fn cmp(&self, other: &Self) -> cmp::Ordering {
 		other.score.cmp(&self.score)
-			.then(other.transaction.insertion_id.cmp(&self.transaction.insertion_id))
+			.then(self.transaction.insertion_id.cmp(&other.transaction.insertion_id))
 	}
 }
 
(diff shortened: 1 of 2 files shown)


transaction-pool/src/scoring.rs
```diff
@@ -138,7 +138,7 @@ impl<T, S: Clone> Clone for ScoreWithRef<T, S> {
 impl<S: cmp::Ord, T> Ord for ScoreWithRef<T, S> {
 	fn cmp(&self, other: &Self) -> cmp::Ordering {
 		other.score.cmp(&self.score)
-			.then(other.transaction.insertion_id.cmp(&self.transaction.insertion_id))
+			.then(self.transaction.insertion_id.cmp(&other.transaction.insertion_id))
```
dvdplm: Am I reading this right? First compare the scores; if equal then compare the insertion_ids and pick the lower (older?) one. That way in a no-gas scenario txs are processed oldest-first? Maybe a comment is warranted here?

tomusdrw: Correct, the tests are documenting that behaviour.

transaction-pool/src/tests/mod.rs
```diff
@@ -210,21 +210,24 @@ fn should_construct_pending() {
 
 	let tx0 = txq.import(b.tx().nonce(0).gas_price(5).new()).unwrap();
 	let tx1 = txq.import(b.tx().nonce(1).gas_price(5).new()).unwrap();
+
+	let tx9 = txq.import(b.tx().sender(2).nonce(0).new()).unwrap();
```
andresilva: Why did you need to change the order of imported of txs?

tomusdrw: Cause it was actually affected by the change - previously `tx9` was returned as one of the first pending transactions, because it was inserted last (with gas_price=0 as the rest).

I didn't change the `assert`, but rather changed the ordering, to make sure the tests are passing.

andresilva: Thanks :+1: Just thought ordering wasn't affected on all of them but I probably overlooked the test.

PR #44, opened 2018-08-24
Link to paritytech.io in README.md

pepyakin:


dvdplm:
The test failures here is because the `fixed-hash` crate tests need the `uint_conversions` feature. If you rebase it will be fine I should think.

PR #46, opened 2018-08-27
kvdb, replace elastic-array with Vecs

tafia:
Following #29 and #40, replaces ElasticArray with Vecs for kvdb, kvdb-memorydb and kvdb-rocksdb.

There are no benchmarks to check the impacts.

### DIFF
--- a/kvdb-memorydb/src/lib.rs
+++ b/kvdb-memorydb/src/lib.rs
@@ -101,7 +101,7 @@ impl KeyValueDB<DBKey, DBValue> for InMemory {
 	}
 
 	fn iter_from_prefix<'a>(&'a self, col: Option<u32>, prefix: &'a [u8])
-		-> Box<Iterator<Item=(DBKey, DBValue)> + 'a>
+		-> Box<Iterator<Item = (DBKey, DBValue)> + 'a>
 	{
 		match self.columns.read().get(&col) {
 			Some(map) => Box::new(
--- a/kvdb-memorydb/src/lib.rs
+++ b/kvdb-memorydb/src/lib.rs
@@ -91,7 +91,7 @@ impl KeyValueDB<DBKey, DBValue> for InMemory {
 		Ok(())
 	}
 
-	fn iter<'a>(&'a self, col: Option<u32>) -> Box<Iterator<Item=(DBKey, DBValue)> + 'a> {
+	fn iter<'a>(&'a self, col: Option<u32>) -> Box<Iterator<Item = (DBKey, DBValue)> + 'a> {
 		match self.columns.read().get(&col) {
 			Some(map) => Box::new( // TODO: worth optimizing at all?
 				map.clone().into_iter()
--- a/kvdb-memorydb/src/lib.rs
+++ b/kvdb-memorydb/src/lib.rs
@@ -19,14 +19,18 @@ extern crate kvdb;
 
 use std::collections::{BTreeMap, HashMap};
 use std::io;
+use std::path::Path;
 use parking_lot::RwLock;
-use kvdb::{DBValue, DBTransaction, KeyValueDB, DBOp};
+use kvdb::{DBTransaction, KeyValueDB, DBOp};
+
+pub type DBValue = Vec<u8>;
+pub type DBKey = Vec<u8>;
 
 /// A key-value database fulfilling the `KeyValueDB` trait, living in memory.
 /// This is generally intended for tests and is not particularly optimized.
 #[derive(Default)]
 pub struct InMemory {
-	columns: RwLock<HashMap<Option<u32>, BTreeMap<Vec<u8>, DBValue>>>,
+	columns: RwLock<HashMap<Option<u32>, BTreeMap<DBKey, DBValue>>>,
 }
 
 /// Create an in-memory database with the given number of columns.
@@ -44,7 +48,7 @@ pub fn create(num_cols: u32) -> InMemory {
 	}
 }
 
-impl KeyValueDB for InMemory {
+impl KeyValueDB<DBKey, DBValue> for InMemory {
 	fn get(&self, col: Option<u32>, key: &[u8]) -> io::Result<Option<DBValue>> {
 		let columns = self.columns.read();
 		match columns.get(&col) {
@@ -53,25 +57,25 @@ impl KeyValueDB for InMemory {
 		}
 	}
 
-	fn get_by_prefix(&self, col: Option<u32>, prefix: &[u8]) -> Option<Box<[u8]>> {
+	fn get_by_prefix(&self, col: Option<u32>, prefix: &[u8]) -> Option<DBValue> {
 		let columns = self.columns.read();
 		match columns.get(&col) {
 			None => None,
-			Some(map) =>
-				map.iter()
-					.find(|&(ref k ,_)| k.starts_with(prefix))
-					.map(|(_, v)| v.to_vec().into_boxed_slice())
+			Some(map) => map.iter()
+				.find(|&(ref k ,_)| k.starts_with(prefix))
+				.map(|(_, v)| v)
+				.cloned(),
 		}
 	}
 
-	fn write_buffered(&self, transaction: DBTransaction) {
+	fn write_buffered(&self, transaction: DBTransaction<DBKey, DBValue>) {
 		let mut columns = self.columns.write();
 		let ops = transaction.ops;
 		for op in ops {
 			match op {
 				DBOp::Insert { col, key, value } => {
 					if let Some(col) = columns.get_mut(&col) {
-						col.insert(key.into_vec(), value);
+						col.insert(key, value);
 					}
 				},
 				DBOp::Delete { col, key } => {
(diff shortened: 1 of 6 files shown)


kvdb/src/lib.rs
```diff
@@ -70,48 +62,36 @@ impl DBOp {
...
 		}
 	}
 
 	/// Insert a key-value pair in the transaction. Any existing value will be overwritten upon write.
-	pub fn put(&mut self, col: Option<u32>, key: &[u8], value: &[u8]) {
-		let mut ekey = ElasticArray32::new();
-		ekey.append_slice(key);
+	pub fn put<IK, IV>(&mut self, col: Option<u32>, key: IK, value: IV)
+	where IK: Into<K>,
+		  IV: Into<V>,
```
pepyakin: There are some spaces. Could you please change them to tabs?

tafia: done

dvdplm:
@tafia can you sort out the conflicts here to please so we can get this one in too? :D

tafia:
done! I'll fix it if there is an issue with travis.

tafia:
All good

kvdb/src/lib.rs
```diff
@@ -134,21 +115,22 @@ impl DBTransaction {
...
-pub trait KeyValueDB: Sync + Send {
+pub trait KeyValueDB<K: AsRef<[u8]>, V>: Sync + Send {
+
 	/// Helper to create a new transaction.
-	fn transaction(&self) -> DBTransaction { DBTransaction::new() }
+	fn transaction(&self) -> DBTransaction<K, V> { DBTransaction::new() }
 
 	/// Get a value by key.
-	fn get(&self, col: Option<u32>, key: &[u8]) -> io::Result<Option<DBValue>>;
+	fn get(&self, col: Option<u32>, key: &[u8]) -> io::Result<Option<V>>;
```
andresilva: Would it make sense to change this to `key: K`? Same for `get_by_prefix` and `iter_from_prefix`?

kvdb-memorydb/src/lib.rs
```diff
@@ -87,32 +91,28 @@ impl KeyValueDB for InMemory {
...
 
 	fn iter_from_prefix<'a>(&'a self, col: Option<u32>, prefix: &'a [u8])
-		-> Box<Iterator<Item=(Box<[u8]>, Box<[u8]>)> + 'a>
+		-> Box<Iterator<Item=(DBKey, DBValue)> + 'a>
 	{
 		match self.columns.read().get(&col) {
 			Some(map) => Box::new(
-				map.clone()
-					.into_iter()
+				map.clone().into_iter()
```
andresilva:
```
map.clone()
	.into_iter()
...
```

kvdb-rocksdb/src/lib.rs
```diff
@@ -563,27 +546,30 @@ impl Database {
...
 		}
 	}
 
-	fn iter_from_prefix(&self, col: Option<u32>, prefix: &[u8]) -> Option<DatabaseIterator> {
+	fn iter_from_prefix<'a>(
+		&'a self,
+		col: Option<u32>,
+		prefix: &[u8]
+	) -> Option<impl Iterator<Item=(DBKey, DBValue)> + 'a>
+	{
```
andresilva: I would place this bracket on the previous line.

kvdb-rocksdb/src/lib.rs
```diff
@@ -563,27 +546,30 @@ impl Database {
...
-			},
+				let iter = col.map_or_else(
+						|| db.iterator_opt(IteratorMode::From(prefix, Direction::Forward), &self.read_opts),
+						|c| db.iterator_cf_opt(
+							cfs[c as usize],
+							IteratorMode::From(prefix, Direction::Forward),
+							&self.read_opts,
+						).expect("iterator params are valid; qed"));
+
+				Some(interleave_ordered(Vec::new(), iter.map(|(k, v)| (k.into(), v.into()))))
```
andresilva: Since we're returning `impl Iterator` we don't need to interleave with an empty `Vec` anymore right? We should be able to just return `iter`.

kvdb/src/lib.rs
```diff
@@ -70,48 +62,37 @@ impl DBOp {
...
 		DBTransaction {
 			ops: Vec::with_capacity(cap)
 		}
 	}
 
 	/// Insert a key-value pair in the transaction. Any existing value will be overwritten upon write.
-	pub fn put(&mut self, col: Option<u32>, key: &[u8], value: &[u8]) {
-		let mut ekey = ElasticArray32::new();
-		ekey.append_slice(key);
+	pub fn put<IK, IV>(&mut self, col: Option<u32>, key: IK, value: IV)
```
andresilva: Is defining `IK` and `IV` with into conversions useful here?

kvdb-rocksdb/src/lib.rs
tafia: Probably ...
I am not sure I understand the reason for the interleaved-ordered in the master code. I guess it is to be as close as `iter` as possible?
I'll try removing it.

kvdb/src/lib.rs
tafia: I believe it is better than without.

Just to be sure, if you mean:
* we should directly have `K` and `V`, then this is for compatibility reason (less code change when migrating to this lib, `&[u8]` will probably work for most cases).
* we should keep only `&[u8]` everywhere, then the `Into` allows for more efficient code (one can transfer ownership of some `K` or `V` directly?)

Note that there may be even better ways.

kvdb/src/lib.rs
tafia: `&K` for the `get` is probably ok (I'll try it). I am unsure for the `*_by_prefix` because this is a *partial key*.

tafia:
Thanks for the review @andresilva. I have made most of the changes you've requested (and answered the other ones). Let me know.

sorpaas:
@tafia Needs conflict resolve!

kvdb-memorydb/src/lib.rs
```diff
@@ -44,34 +48,34 @@ pub fn create(num_cols: u32) -> InMemory {
...
 			None => None,
-			Some(map) =>
-				map.iter()
-					.find(|&(ref k ,_)| k.starts_with(prefix))
-					.map(|(_, v)| v.to_vec().into_boxed_slice())
+			Some(map) => map.iter()
+				.find(|&(ref k ,_)| k.starts_with(prefix))
+				.map(|(_, v)| v)
+				.cloned(),
 		}
```
niklasad1: Not introduced by this PR but pattern matching could be replaced by `and_then` to eliminate `None => None` etc

```rust
		self.columns.read().get(&col)
			.and_then(|db| {
				db.iter()
					.find(|&(ref k ,_)| k.starts_with(prefix))
					.map(|(_, v)| v)
					.cloned()
			})

```

kvdb-memorydb/src/lib.rs
```diff
@@ -87,32 +91,29 @@ impl KeyValueDB for InMemory {
 		Ok(())
 	}
 
-	fn iter<'a>(&'a self, col: Option<u32>) -> Box<Iterator<Item=(Box<[u8]>, Box<[u8]>)> + 'a> {
+	fn iter<'a>(&'a self, col: Option<u32>) -> Box<Iterator<Item=(DBKey, DBValue)> + 'a> {
```
niklasad1: style

```suggestion
	fn iter<'a>(&'a self, col: Option<u32>) -> Box<Iterator<Item = (DBKey, DBValue)> + 'a> {
```

kvdb-memorydb/src/lib.rs
```diff
@@ -87,32 +91,29 @@ impl KeyValueDB for InMemory {
...
-					.map(|(k, v)| (k.into_boxed_slice(), v.into_vec().into_boxed_slice()))
+				map.clone().into_iter()
 			),
 			None => Box::new(None.into_iter()),
 		}
 	}
 
 	fn iter_from_prefix<'a>(&'a self, col: Option<u32>, prefix: &'a [u8])
-		-> Box<Iterator<Item=(Box<[u8]>, Box<[u8]>)> + 'a>
+		-> Box<Iterator<Item=(DBKey, DBValue)> + 'a>
```
niklasad1: style 

```suggestion
		-> Box<Iterator<Item = (DBKey, DBValue)> + 'a>
```

kvdb-rocksdb/src/lib.rs
```diff
@@ -545,14 +529,13 @@ impl Database {
 	}
 
 	/// Get database iterator for flushed data.
-	pub fn iter(&self, col: Option<u32>) -> Option<DatabaseIterator> {
+	pub fn iter<'a>(&'a self, col: Option<u32>) -> Option<impl Iterator<Item=(DBKey, DBValue)> + 'a> {
```
niklasad1: style 

```suggestion
	pub fn iter<'a>(&'a self, col: Option<u32>) -> Option<impl Iterator<Item = (DBKey, DBValue)> + 'a> {
```

kvdb-rocksdb/src/lib.rs
```diff
@@ -563,27 +546,29 @@ impl Database {
...
 			None => None,
 		}
 	}
 
-	fn iter_from_prefix(&self, col: Option<u32>, prefix: &[u8]) -> Option<DatabaseIterator> {
+	fn iter_from_prefix<'a>(
+		&'a self,
+		col: Option<u32>,
+		prefix: &[u8]
+	) -> Option<impl Iterator<Item=(DBKey, DBValue)> + 'a> {
```
niklasad1:
```suggestion
	) -> Option<impl Iterator<Item = (DBKey, DBValue)> + 'a> {
```

dvdplm:
> There are no benchmarks to check the impacts.

The reason this PR has not been merged yet is that it has a potentially big impact upstream. Therefore I think it's important to perform the above benchmarking in parity-ethereum. The code changes per se lgtm.

sorpaas:
@dvdplm I think it's generally safe -- we had past experience that replacing elastic array with vec improved performance. To simplify parity-ethereum testing, maybe we can merge this first, and then PR benchmark on parity-ethereum, if things go wrong, revert it? The issue right now is that we cannot directly benchmark with only this repo.

dvdplm:
> maybe we can merge this first, and then PR benchmark on parity-ethereum,

How about releasing a `-beta` crate with the unmerged code? The reason I never go around to benchmarking this is there wasn't any good high-level benchmarks I trusted enough to run. These changes assume `Vec` is faster than `ElasticArray` and while I do believe that to be true, I also think the performance is dependent on the workload. The benchmark we use to prove that it is an improvement should use realistic workloads imo. If we have such benchmarks now, then by all means, let's run them! :)

tafia:
> These changes assume Vec is faster than ElasticArray and while I do believe that to be true, I also think the performance is dependent on the workload

Just to make it clear, this is less about having performance wins than removing some unsafes. I am just expecting not having worse perf, which indeed need to be tested somehow.

Having realistic benchmarks is definitely that needs to be done. I am unfortunately not expert enough to know what realistic means.

Sorry for the late answers!

kvdb/src/lib.rs
```diff
@@ -134,21 +115,22 @@ impl DBTransaction {
 ///
 /// The API laid out here, along with the `Sync` bound implies interior synchronization for
 /// implementation.
-pub trait KeyValueDB: Sync + Send {
+pub trait KeyValueDB<K: AsRef<[u8]>, V>: Sync + Send {
```
dvdplm: I am having doubts about how this is actually works out in consuming code. As it stands before this PR, consuming code can use different key types freely. For example in `parity-ethereum` we use `[u8; 12]`, `H256` and `&[u8]` (and possibly others) without any particular gymnastics required. With this PR and `K` set to be `Vec<u8>` all kinds of "fun" conversions become necessary for little apparent gain.

I know this is a lot to ask @tafia but could you take a look at for example `journaldb` ([here](https://github.com/paritytech/parity-ethereum/tree/master/util/journaldb)) and show me how you'd convert it to use this PR? When I try it gets painful fast.

PR #68, opened 2018-09-30
Add missing repository tags

pepyakin:


hashdb/Cargo.toml
```diff
@@ -3,6 +3,7 @@ name = "hashdb"
 version = "0.3.0"
 authors = ["Parity Technologies <admin@parity.io>"]
 description = "trait for hash-keyed databases."
+repository = "https://github.com/paritytech/parity-common"
```
tomusdrw: Should we have homepage there as well?

dvdplm: IMO only if we actually have a homepage (and we don't).

PR #69, opened 2018-10-04
Add optional serialization feature via serde

Robbepop:
This change is required in order to provide a fluent way for `fleetwood` (and its dependency: `pwasm-abi`) to update and make `fleetwood` compile again. Also it is handy to have optional serde support.

The new feature is named `serialize` just as it was called in the deprecated `bigint` crate although `serde-1` might be a better choice.

### DIFF
--- a/uint/src/uint.rs
+++ b/uint/src/uint.rs
@@ -350,6 +350,7 @@ macro_rules! construct_uint {
 		/// Little-endian large integer type
 		#[repr(C)]
 		#[derive(Copy, Clone, Eq, PartialEq, Hash)]
+		#[cfg_attr(feature="serialize", derive(Serialize, Deserialize))]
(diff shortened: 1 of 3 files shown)


uint/src/uint.rs
```diff
@@ -350,6 +350,7 @@ macro_rules! construct_uint {
 		/// Little-endian large integer type
 		#[repr(C)]
 		#[derive(Copy, Clone, Eq, PartialEq, Hash)]
+		#[cfg_attr(feature="serialize", derive(Serialize, Deserialize))]
 		pub struct $name(pub [u64; $n_words]);
```
debris: this deserializes the value as a list which is unacceptable

debris:
I'm closing as this change will not be accepted

Robbepop:
Okay thank you very much for the pointers and links!

PR #76, opened 2018-10-31
[WIP] Fixed hash 0.3 beta fixes

Robbepop:
Fixes for the beta of the `0.3` release.

Fixes so far:

- Optional crate features of `fixed-hash` (e.g. `byteorder`, formerly known as `byteorder-support`) no longer leak into user crates. Done by https://github.com/paritytech/parity-common/pull/76/commits/1f94d2747d85c7650534f2b595d908a498d936b6.
- Crate `core` is now publicly exported as `core_`
    - Fixes a bug that `fixed-hash` 0.3.0-beta.0 wasn't usable in `no_std` environments
- Macro `const_assert` is now publicly exported (hidden docs)
    - This has the effect that users no longer need to import `static_assertions` themselves to have this macro in scope
- Add `Hash::to_bytes` method which is especially useful for efficient conversions between `fixed-hash` and `uint`.

**NOTE: THIS IS WORK IN PROGRESS! DO NOT MERGE!**

### DIFF
--- a/fixed-hash/src/hash.rs
+++ b/fixed-hash/src/hash.rs
@@ -292,6 +292,7 @@ macro_rules! construct_fixed_hash {
 		{
 			type Output = I::Output;
 
+			#[inline]
 			fn index(&self, index: I) -> &I::Output {
 				&self.as_bytes()[index]
 			}
@@ -301,6 +302,7 @@ macro_rules! construct_fixed_hash {
 		where
 			I: $crate::core_::slice::SliceIndex<[u8], Output = [u8]>
 		{
+			#[inline]
 			fn index_mut(&mut self, index: I) -> &mut I::Output {
 				&mut self.as_bytes_mut()[index]
 			}
@@ -750,7 +752,6 @@ macro_rules! impl_ops_for_hash {
 		$ops_tok:tt,
 		$ops_assign_tok:tt
 	) => {
-
 		impl<'r> $crate::core_::ops::$ops_assign_trait_name<&'r $impl_for> for $impl_for {
 			fn $ops_assign_fn_name(&mut self, rhs: &'r $impl_for) {
 				for (lhs, rhs) in self.as_bytes_mut().iter_mut().zip(rhs.as_bytes()) {
@@ -760,6 +761,7 @@ macro_rules! impl_ops_for_hash {
 		}
 
 		impl $crate::core_::ops::$ops_assign_trait_name<$impl_for> for $impl_for {
+			#[inline]
 			fn $ops_assign_fn_name(&mut self, rhs: $impl_for) {
 				*self $ops_assign_tok &rhs;
 			}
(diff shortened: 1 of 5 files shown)


fixed-hash/src/lib.rs
```diff
@@ -8,6 +8,11 @@
 
 #![cfg_attr(not(feature = "std"), no_std)]
 
+// Re-export libcore using an alias so that the macros can work without
+// requiring `extern crate core` downstream.
+#[doc(hidden)]
+pub extern crate core as core_;
```
dvdplm: Hmm. This is somewhat unsatisfying. Isn't it normal for a `no_std` crate to add `core` (if they need it). With this a downstream crate will add `core` implicitly which is perhaps undesired? 
What is the alternative to this, if any?

Robbepop: As far as @pepyakin and me found out this is the general workaround for the problem.
Also used in other crates, such as `static_assertions`.

Robbepop:
Since the inner field of all hash types is private now we need more ways to get views into the data.
Proposed new APIs:
```rust
fn as_fixed_bytes(&self) -> &[u8; N];
fn as_fixed_bytes_mut(&mut self) -> &[u8; N];
```
Also rename
```rust
fn to_bytes(self) -> [u8; N];
```
to
```rust
fn to_fixed_bytes(self) -> [u8; N];
```
for regularity.

Robbepop:
This has had time to forge to a new stable interface and I am going to release a version `0.3.0` (no more beta) now.

Robbepop:
Just released version `0.3` of `fixed-hash` on crates.io.
I would really like to merge this PR now.
Need some more review approvials for that.

fixed-hash/src/hash.rs
```diff
@@ -485,9 +535,110 @@ macro_rules! construct_fixed_hash {
...
+// 
+// # Note
+//
+// Feature guarded macro definitions instead of feature guarded impl blocks
+// to work around the problems of introducing `libc` crate feature in
+// a user crate.
+#[cfg(all(feature = "libc", not(target_os = "unknown")))]
+#[macro_export]
+#[doc(hidden)]
+macro_rules! impl_libc_for_fixed_hash {
```
dvdplm: I still have not been able to understand what advantage libc support actually brings. Is it more performant to impl `PartialEq`/`Ord` like this?

Robbepop: I will just merge this now and we can resolve this issue in a later PR, is that okay?

dvdplm: Sure, just open an issue for it please?

pepyakin: Personally, I don't want to block merging this PR because of this.

Robbepop: will do

Robbepop: done: https://github.com/paritytech/parity-common/issues/79

Robbepop: merging now

PR #80, opened 2018-11-15
Secp256k1 in parity crypto and alternative wasm32 implementation

cheme:
# Removal of rust-crypto

This PR superseed #14 , also probably nice to have for https://github.com/paritytech/parity-ethereum/issues/7915

RustCrypto hashing crate are already used Parity-ethereum (deps from new hyper version).

AES, on the other hand is new, possible change:
- adopt a conservative approach and restore old unmaintained AES rust-crypto implementation and put RustCrypto in an alt implementation crate.

Also note that we can enable aes-ni while building for a nice bench boost (requires `RUSTFLAGS="-C target-feature=+aes,+sse2,+ssse3" cargo bench`) :

```
aes_ctr/10000           time:   [3.5767 us 3.5778 us 3.5789 us]                           
                        change: [-98.174% -98.174% -98.173%] (p = 0.00 < 0.05)
                        Performance has improved.
  3 (3.00%) high mild
aes_ctr/100000          time:   [36.707 us 36.740 us 36.778 us]                            
                        change: [-98.091% -98.089% -98.087%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 14 outliers among 100 measurements (14.00%)
```

# Alternate implementation

This PR exposes alternate implementation (mainly for wasm32 usecase), the choice of crates is less secure but it is an alternate implementation (wasm32 does not give the same garantie as native code).

Possible change:
- add feature switch to make them default implementation (for custom build for instance)

# secp256k1

## basic integration

A quick and dirty basic integration was running correctly at this point (really conservative) : https://github.com/cheme/parity-common/blob/wasm-test/parity-crypto/src/secp256k1.rs and https://github.com/cheme/parity-common/blob/wasm-test/parity-crypto/src/secp256k1_alt.rs .

This can still be an alternative but is not really fit for a common crate.

## trait

Using the previous basic integration reveals some small discrepancy between both crates.
At the end it seems obvious that using a trait to reduce exposed functionality makes senses to ensure compatibility in the long term.

A trait was define for secp in module traits.

Some alternative:
- At a some point I made `PublicKey` and `SecretKey` implements `AsRef<u8>`, but I revert back (if we want to maintain a cache of serialize value it should be at parity-ethereum/ethkey instead).
- If reverting back to being `AsRef<u8>` use `fixed-hash` for inner cache, using H512 to get efficient equality, and derive ordering, probably not it should be done at ethkey level.
- define `Signature` as a trait, I see signature as a 'use once' operation so I keept things light with bytes here.
- Public key size for secp to 65 (currently 64 without uncompressed tag) could be more correct.
- Add default value to PublicKey : this was done at some point but feels wrong.
- error changes the way debug trait behave, originally the native error where displayed in debug mode from json rpc therefore parity ethereum rpc changes a bit. I only identified one message, so I do not think a compatibility layer is needed, but that is an open question.

## Secret Store

On my parity branch (https://github.com/cheme/parity-ethereum/tree/crypto-only), secret store switch to the secp trait did not work well due to a few design point:
- default public key use in test (and asymption on its first position in btree), the value is an invalid publickey that cannot be spawn.

Currently on branch there is a mix of `NodeId` aka `H512` and actual full `PublicKey`, with a lot of conversion between each. Moreover there is 
a artificial unsafe `Default` implementation in secp256k1.rs.

Possibility:
- do not switch secret-store : keep dependancy on secp crate for secret-store only
- revert to a more conservative approach : use `NodeId` (H512) everywhere and `Public` only for spawning and maths
- Integrate Public everywhere (except `Serialize` objects) by very involved refactoring
- keep as is on the branch, even if tests are passing I am not to confident with that, and there is already a lot of byte copy involved.

## bench

I initiated some bench, but there is not much.
`bench --feature alt` to get both secp implementation benched (alt tests are run without this feature it is related to an issue with bench not running with deps embedding test code).

## clear on drop

crate clear_on_drop is used and exposed, MemZero from parity-ethereum has been copied to parity common and can both use clean_on_drop or previous implementation.

Possibility:
- I realize while creating this pr that I use internally ClearOnDrop instead of MemZero, it could make sense to use MemZero

## Next steps

I did fail at keeping the PR small (especially when considering https://github.com/cheme/parity-ethereum/tree/crypto-only).
Next step is integration into parity-ethereum (after pushing crate version of course), then others PR to use trait for other crypto things. Doing things in a iterative way.

Another important thing is to make the crate `no_std` compatible (at least the traits and some implementations).

Also a TODO in secp alt impl for clearing memory (we need an unsafe method to access inner field of secret key).

Some alternative:
- wait for better math design (to replace ethstore and ethkey secret store impls)
- wait for other traits impl.


parity-crypto/src/scrypt.rs
```diff
@@ -15,8 +15,10 @@
 // along with Parity.  If not, see <http://www.gnu.org/licenses/>.
 
 use error::ScryptError;
-use rcrypto::scrypt::{scrypt, ScryptParams};
 use super::{KEY_LENGTH_AES, KEY_LENGTH};
+use rscrypt:: {scrypt, ScryptParams};
```
sorpaas: nit: extra whitespace

parity-crypto/src/scrypt.rs
```diff
@@ -15,8 +15,10 @@
...
+#[cfg(test)]
+use std::io::{ Error };
```
sorpaas: nit: unneeded bracket

parity-crypto/src/secp256k1_alt.rs
```diff
@@ -0,0 +1,411 @@
...
+	type VecRepr = ClearOnDrop<Vec<u8>>;
+
+	fn to_vec(&self) -> Self::VecRepr {
+		ClearOnDrop::new(self.0.serialize().to_vec())
+	}
+
+	fn sign(&self, message: &[u8]) -> Result<Vec<u8>, Error> {
+
+	 	let mut buf = [0;32];
+		buf.copy_from_slice(&message[..]); // panic on incorrect message size
```
sorpaas: > panic on incorrect message size

This looks like a TODO. Should we check and return error?

parity-crypto/src/secp256k1_alt.rs
```diff
@@ -0,0 +1,411 @@
...
+}
+*/
+
+impl FiniteField for Secp256k1 {
+
+	fn generator_x() -> &'static[u8] { &GENERATOR_X[..] }
+	fn generator_y() -> &'static[u8] { &GENERATOR_Y[..] }
+	fn curve_order() -> &'static[u8] { &CURVE_ORDER[..] }
+
+	fn public_mul(pub_key: &mut Self::PublicKey, sec_key: &Self::SecretKey) -> Result<(), Error> {
```
sorpaas: Want to ask -- what are the use cases for `public_mul` / `public_add` / `secret_add` / `secret_mul` / `secret_inv`? If they're useful, I should implement them in `libsecp256k1`.

parity-crypto/src/secp256k1_alt.rs
```diff
@@ -0,0 +1,411 @@
...
+		let mut data_arr = vec![0; 65];
+		data_arr[0..64].copy_from_slice(&sig.serialize());
+		data_arr[64] = rec_id.serialize();
+		Ok(data_arr)
+	}
+
+}
+
+/// random secret key for rand 0.5
+fn random_sec<R: Rng>(rng: &mut R) -> SecretKey {
```
sorpaas: Does `SecretKey::random` work for this purpose?

parity-crypto/src/secp256k1_alt.rs
cheme: It is mainly here for compatibility (the point of the crate is not the clean api/trait), so it is operation on key use by :
- secret_store for key calculation/distribution/verification.
- ethkey for all key derivation operations (hd wallet).

It would be certainly better if the lib got higher level operation (moving those low level key math (there is probably some redundancy between each crates math primitives) to a single parity crypto impl.

But to keep thing simple I use those trait.

parity-crypto/src/secp256k1_alt.rs
cheme: I needed external rng source for wasm32 (not using emscripten), but this is not good either the dependency on Rng trait can be avoid through only exposing `secret_from_slice`. I think I should totally remove `generate_keypair` and this function.

parity-crypto/src/secp256k1_alt.rs
sorpaas: Got this as `PublicKey::tweak_add_assign`.

parity-crypto/src/secp256k1_alt.rs
```diff
@@ -0,0 +1,411 @@
...
+		let mut zero = Scalar::default();
+		zero.set_int(0);
+		let mut res = Jacobian::default();
+		ECMULT_CONTEXT.ecmult(&mut res, &pub_jac, &sec_scal.0, &zero);
+		pub_aff.set_gej(&res);
+		*pub_key = PublicKey::new(aff_to_public(&mut pub_aff)?);
+		Ok(())
+	}
+
+	fn public_add(pub_key: &mut Self::PublicKey, other_public: &Self::PublicKey) -> Result<(), Error> {
```
sorpaas: Got this as `PublicKey::tweak_add_assign`.

sorpaas: Add this in `libsecp256k1` as `PublicKey::combine`.

parity-crypto/src/secp256k1_alt.rs
```diff
@@ -0,0 +1,411 @@
...
+	/// Warning we use 64 bit pubsize (first bytes of 65 bit representation is 4).
+	const PUB_SIZE: usize = PUB_SIZE;
+
+	const SECRET_SIZE: usize = SECRET_SIZE;
+
+	const KEYPAIR_INPUT_SIZE: usize = Self::SECRET_SIZE;
+
+	fn recover(signature: &[u8], message: &[u8]) -> Result<Self::PublicKey, Error> {
+		let mut buf = [0;32];
+		buf.copy_from_slice(&message[..]); // panic on incorrect message size
```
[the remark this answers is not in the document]
cheme: Ok I am seeing those commits on your crate, I will use them, it is probably best if I wait a bit, thanks :+1: 

I also got a thing I couldn't do : 
https://github.com/cheme/parity-common/blob/30d91988378d0f0c4c070c01620107e65334950a/parity-crypto/src/secp256k1_alt.rs#L123-L130
Basically if there was an 'unsafe' accessor to the secret key 'Field' I could use Clear on it.

parity-crypto/src/secp256k1_alt.rs
sorpaas: Added this as `PublicKey::tweak_mul_assign`.

parity-crypto/src/secp256k1_alt.rs
```diff
@@ -0,0 +1,411 @@
...
+		aff_pub_j.set_ge(&aff_pub);
+		let aff_pub_other: Affine = other_public.0.clone().into();
+		let res_j = aff_pub_j.add_ge(&aff_pub_other);
+		aff_pub.set_gej(&res_j);
+		*pub_key = PublicKey::new(aff_to_public(&mut aff_pub)?);
+
+		Ok(())
+	}
+
+	fn secret_mul(sec_key: &mut Self::SecretKey, other_sec_key: &Self::SecretKey) -> Result<(), Error> {
```
sorpaas: As `SecretKey::tweak_mul_assign`

parity-crypto/src/secp256k1_alt.rs
```diff
@@ -0,0 +1,411 @@
...
+		let sec_scal = SecretScalar(sec_key.0.clone().into());
+		let other_sec_scal = SecretScalar(other_sec_key.0.clone().into());
+		// we could use * operator instead.
+		let mut res = SecretScalar(Scalar::default());
+		res.0.mul_in_place(&sec_scal.0, &other_sec_scal.0);
+		*sec_key = SecretKey::new(SecretKeyInner::parse(&res.0.b32())?);
+		Ok(())
+	}
+
+	fn secret_add(sec_key: &mut Self::SecretKey, other_sec_key: &Self::SecretKey) -> Result<(), Error> {
```
sorpaas: As `SecretKey::tweak_add_assign`

parity-crypto/src/secp256k1_alt.rs
```diff
@@ -0,0 +1,411 @@
...
+		let sec_scal = SecretScalar(sec_key.0.clone().into());
+		let other_sec_scal = SecretScalar(other_sec_key.0.clone().into());
+		// we could use + operator instead.
+		let mut res = SecretScalar(Scalar::default());
+		res.0.add_in_place(&sec_scal.0, &other_sec_scal.0);
+		*sec_key = SecretKey::new(SecretKeyInner::parse(&res.0.b32())?);
+		Ok(())
+	}
+
+	fn secret_inv(sec_key: &mut Self::SecretKey) -> Result<(), Error> {
```
sorpaas: Added `SecretKey::inv`.

parity-crypto/src/secp256k1_alt.rs
dvdplm: > It is mainly here for compatibility

Maybe this PR is a good chance to add a deprecation notice?

parity-crypto/src/secp256k1_alt.rs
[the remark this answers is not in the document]
sorpaas: Will adding `Default` trait to `SecretKey` work? It looks like in this way it automatically implements `cod::clear::Clear`: https://github.com/sorpaas/libsecp256k1-rs/blob/master/tests/verify.rs#L53

I currently set the "default" secret key to `0, 0, ..., 1` because all zero is not valid.

sorpaas:
Published `libsecp256k1 0.2` so that we can test this without affecting existing users (if any!).

parity-crypto/src/secp256k1_alt.rs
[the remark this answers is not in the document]
cheme: It works (I wish I could do it in this crate to avoid it but my tests where unsuccessfull).

parity-crypto/src/secp256k1_alt.rs
cheme: My tests are failing with combine, I think it's related to an error when manipulating infinite point for first addition.

parity-crypto/src/secp256k1_alt.rs
cheme: Would probably make sense in a second step; seems a bit premature now (moving those math usage is a bit of a refactoring and I am not sure I can assert that it is doable or just even wanted at this point).

parity-crypto/src/secp256k1_alt.rs
sorpaas: It's a bug in `add_ge`. Sorry! Fixed and published `0.2.1`.

parity-crypto/src/aes.rs
```diff
@@ -14,40 +14,182 @@
...
+
+
+// retrocomptibility test
+#[test]
+pub fn test_aes_short() -> Result<(),SymmError> {
+	let key = include_bytes!("../test/key1");
+	let salt = include_bytes!("../test/salt1");
+	let content = include_bytes!("../test/content");
+	let ctr_enc = include_bytes!("../test/result_128_ctr");
+	let cbc_enc = include_bytes!("../test/result_128_cbc");
```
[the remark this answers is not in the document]
cheme: Will do.

parity-crypto/src/hmac.rs
```diff
@@ -86,3 +86,51 @@ impl VerifyKey<digest::Sha512> {
...
+		let sig_key1 = SigKey::sha256(&key1[..]);
+		let sig_key2 = SigKey::sha512(&key2[..]);
+
+		let mut signer1 = Signer::with(&sig_key1);
+		let mut signer2 = Signer::with(&sig_key2);
+
+		signer1.update(&input[..]);
+		for i in 0 .. big_input.len() / 33 {
+			signer2.update(&big_input[i*33..(i+1)*33]);
+		}
```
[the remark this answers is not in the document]
cheme: Just running the update on slices that are not  64 byte multiple, a best test would use different sizes.

mem/src/lib.rs
```diff
@@ -0,0 +1,74 @@
...
+	fn drop(&mut self) {
+		unsafe {
+			for byte_ref in self.mem.as_mut() {
+				ptr::write_volatile(byte_ref, 0)
+			}
+		}
+	}
+}
+
+#[cfg(not(feature = "volatile-erase"))]
```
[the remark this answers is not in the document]
cheme: I did use 'clear_on_drop' in my implementation (it is a bit more flexible api-wise) in other contexts and feel like using a single erasure mechanism was simpler. 
The crate in itself got its own advantages (https://github.com/cesarb/clear_on_drop/issues/2), yet for me it was just a matter of giving the choice (under current configuration it is imported with 'volatile' feature activated in parity-crypto).

cheme:
@twittner  

> Instead of having this single PR, would it be possible to move from 1 to 3 one PR at a time?

Could do, it will be probably in  1-3-2 order. Only thing will be that my current parity-ethereum branch testing this pr will be hard to align, but it do not really have to if I maintain a sync of those three branch with this branch.

mem/src/lib.rs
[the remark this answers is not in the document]
dvdplm: I think in this case "less is more". I think we should have a single way of doing memory zeroing.

mem/Cargo.toml
```diff
@@ -0,0 +1,11 @@
+[package]
+name = "mem"
+version = "0.1.0"
+authors = ["Parity Technologies <admin@parity.io>"]
```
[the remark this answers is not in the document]
dvdplm: Sure. Just make sure you add the required fields to the `[package]` section here (or else it can't be published).

parity-crypto/Cargo.toml
```diff
@@ -5,9 +5,52 @@ authors = ["Parity Technologies <admin@parity.io>"]
...
 tiny-keccak = "1.4"
+scrypt = { version = "0.1.1", default-features = false }
+ripemd160 = "0.8.0"
+sha2 = "0.8.0"
+digest = "0.8"
+aes = "0.3.2"
+aes-ctr = "0.3.0"
+block-modes = "0.2.0"
+lazy_static = "1.0" # for secp
+mem = { path = "../mem" }
```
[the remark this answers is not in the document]
dvdplm: > I am not totally sure that I want to get this crate published 

In that case I think it's best to make it a sub-crate to `parity-crypto`. If in the future we realize it has value outside that context, moving it is easy.

I'd prefer the top-level crates in `parity-common` to all be published.

mem/Cargo.toml
```diff
@@ -0,0 +1,11 @@
+[package]
+name = "mem"
```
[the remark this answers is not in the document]
dvdplm: Hmm, not sure I love that name, sounds a bit like a pathname! How about, as mentioned above, you move it to a `parity-crypto` subcrate and so we publish it as part of that? If you don't like `memzero`, then just `mem-utils` seems like it could work.

dvdplm:
@cheme what is the status of this work? Should we close this in favour of the three split PRs you mention above?

cheme:
I do not really know, the main point of this pr is to be able to compile parity-ethereum to webassembly.
It also go in a different way than recent wasm compatibility pr: it uses feature a lot when recent pr focused on removing non compatible eth implementation. eg aes-gcm
Here we use multiple implementation behind a trait, that is a totally different direction for the crate (this pr see this crate as a unified place for parity related crypto implementation with trait to allow easy inter-operabillity, cross implementation testing, auditability and such but it seems it was not really a goal).
I will still rebase, because while switching from ring to dalek seems good, but switching to other secp for wasm may be another question.
The PR could probably be close afterward, close pr still have the visibility so if we need to get secp to wasm this could be revive easily.

cheme:
also merging this requires a pr on ethereum size that is pretty impacting and maybe not worth it if we do not want to compile to wasm; therefore I am closing the PR.

PR #83, opened 2018-11-21
Documentation fixes

dvdplm:
It's **bytes** not bits. Also updates crunchy to latest.

### DIFF
--- a/plain_hasher/Cargo.toml
+++ b/plain_hasher/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "plain_hasher"
-description = "Hasher for 32-bit keys."
-version = "0.2.0"
+description = "Hasher for 32-byte keys."
+version = "0.2.1"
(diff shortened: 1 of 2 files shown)


plain_hasher/Cargo.toml
```diff
@@ -1,16 +1,16 @@
...
+version = "0.2.1"
 authors = ["Parity Technologies <admin@parity.io>"]
 license = "MIT"
 keywords = ["hash", "hasher"]
 homepage = "https://github.com/paritytech/parity-common"
 categories = [ "no-std"]
 
 [dependencies]
-crunchy = "0.1.6"
+crunchy = "0.2.1"
```
Robbepop: Ah, note that since `crunchy` version `2.1` it is `std` by default. So you require `default-features = false` for it and forward the `std` feature in the `[features]` section for it, too.

Issue #84, opened 2018-11-30
[parity-crypto] Allow easier implementation switch and specific targets


cheme:
This is a meta-issue related to multiple possible improvements :
        - (1) moving all crypto dependency at the same place
        - (2) ability to run a test suite that is directly related to parity use
        - (3) ability to run bench on different crates
        - (4) traits over main parity crypto use, and hability to switch crates without to much efforts (not meant to be std crypto trait, more a convenient way to audit/check our crypto libs usage).
        - (5) wasm32 compatibility/alternative
        - (6) no_std compatibility/alternative

Following steps are proposed :

   - [x] #85 Remove rust-crypto. The rust-crypto crate is unmaintained and in itself is problematics for (5) and (6). Probably going for RustCrypto.
   - [ ] Add support for secp256k1 with traits. Concur to (1) and (4)
   - [ ] Enable cross-compilation to wasm32 target. Concur to (5)
   - [ ] Make the crate no_std compatible. Concur to (6)
   - [ ] Add further trait and tests/bench...
 

A previous PR covered first three steps is at #80 , I put it on ice in favor of smaller PRs. It contains lots of additional information.

PR #85, opened 2018-11-30
rust-crypto dependency removal

cheme:
This PR superseed #14 , is part of #80 split. Related to #84 .

It covers 'rust-crypto' crate removal.

'RustCrypto' crate is used :
- for some hashing function (not really a new dependency for parity eth)
- aes ecb
- aes ctr
The change to ctr is probably the more impacting for parity-ethereum.

I could also adopt a conservative approach and restore old unmaintained AES rust-crypto implementation and put RustCrypto in an alternative implementation crate.

Also worth remembering that we can enable aes-ni while building with RUSTFLAGS="-C target-feature=+aes,+sse2,+ssse3", for improved performances :
```
aes_ctr/10000           time:   [3.5767 us 3.5778 us 3.5789 us]                           
                        change: [-98.174% -98.174% -98.173%] (p = 0.00 < 0.05)
                        Performance has improved.
  3 (3.00%) high mild
aes_ctr/100000          time:   [36.707 us 36.740 us 36.778 us]                            
                        change: [-98.091% -98.089% -98.087%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 14 outliers among 100 measurements (14.00%)
```


parity-crypto/src/hmac.rs
```diff
@@ -86,3 +86,51 @@ impl VerifyKey<digest::Sha512> {
...
 	hmac::verify(&k.0, data, sig).is_ok()
 }
+
+
+#[cfg(test)]
+::tests_hmac!();
+
+#[cfg(test)]
+#[macro_export]
+macro_rules! tests_hmac {
```
sorpaas: It looks like we don't need this macro now?

Also by the way, you can create a separate test file and use `mod` with `path` attribute to include it multiple times. That should make everyone's formatter much happier.

cheme: Thanks for the 'path' siggestion, it will so much cleaner.

tomaka:
Could we revive this PR? It probably needs some master merging.

cheme:
no problem, will be updating that this afternoon :+1:

parity-crypto/src/aes.rs
```diff
@@ -14,40 +14,195 @@
...
+
+	/// New encoder/decoder, no iv for ecb
+	pub fn new(key: &[u8]) -> Result<Self, SymmError> {
+		Ok(AesEcb256(Ecb::new_var(key, &[])?))
+	}
+
+	/// Encrypt data in place without padding. The data length must be a multiple
+	/// of the block size.
+	pub fn encrypt(self, content: &mut [u8]) -> Result<(), SymmError> {
+    let len = content.len();
```
tomaka:
```suggestion
        let len = content.len();
```

tomaka:
I don't know if anyone is watching this PR.
We have two positive reviews, so if anyone wants to review this as well, please express yourself here, otherwise I suggest we merge this in 24 hours.

PR #86, opened 2018-12-02
Unify primitive types for Parity Ethereum and Substrate

sorpaas:
rel https://github.com/paritytech/parity-ethereum/issues/9994

This puts five primitive types into a crate `primitive-types`. They are `U256`, `U512`, `H160`, `H256` and `H512`. This allows us to use parity-ethereum crates on substrate, and vice versa.

* `primitive-types` now implements RLP serialization for the types defined. It's expected that later we remove `ethereum` feature from `rlp` crate.
* Three new helper crate are added -- `impl-serde`, `impl-codec` and `impl-rlp`. This allows code reuse when additional types with similar serialization function need to be defined. It also saves some complications of declaring dependencies.


primitive-types/src/lib.rs
```diff
@@ -0,0 +1,172 @@
...
+// http://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+// <LICENSE-MIT or http://opensource.org/licenses/MIT>, at your
+// option. This file may not be copied, modified, or distributed
+// except according to those terms.
+
+//! Primitive types shared by Substrate and Parity Ethereum.
+
+#![cfg_attr(not(feature = "std"), no_std)]
+
+#[cfg(all(feature = "impl-serde", not(feature = "std")))]
```
dvdplm: Why did you prefer this to `serde = ["std", "impl-serde"]`?

When trying `serde = ["std", "impl-serde"]` locally with `cargo check --no-default-features --features=serde` it seems to build fine?

primitive-types/src/lib.rs
```diff
@@ -0,0 +1,172 @@
...
+#[cfg(feature = "impl-codec")]
+#[macro_use]
+extern crate impl_codec;
+
+#[cfg(feature = "impl-rlp")]
+#[macro_use]
+extern crate impl_rlp;
+
+construct_uint!(U256, 4);
+#[cfg(feature = "impl-serde")] impl_uint_serde!(U256, 4);
```
dvdplm: style nit: put these on two lines

primitive-types/src/lib.rs
```diff
@@ -0,0 +1,172 @@
...
+#[cfg(feature = "impl-codec")] impl_uint_codec!(U256, 4);
+#[cfg(feature = "impl-rlp")] impl_uint_rlp!(U256, 4);
+
+impl U256 {
+	/// Multiplies two 256-bit integers to produce full 512-bit integer
+	/// No overflow possible
+	#[inline(always)]
+	pub fn full_mul(self, other: U256) -> U512 {
```
dvdplm: There is a `common` feature of `uint` that has this already. That feature also builds&exports `U256` and `U512` which imo is a bit of a wart (the idea is that `uint` should only provide the **tools** to construct the types, not the types themselves).
@Robbepop do you remember the reason we added the `common` module/feature`? I'm guessing it was to make it less of a hassle to migrate to the new `uint` version? 

What is stopping us from changing `uint_full_mul_reg!` to implement the all of `pub fn full_mul()`? Can't we do that and then implement it for all types created with `construct_uint!`? That way we can remove this.

primitive-types/src/lib.rs
```diff
@@ -0,0 +1,172 @@
...
+		U512(uint_full_mul_reg!(U256, 4, self, other))
+	}
+}
+
+impl From<U256> for U512 {
```
dvdplm: Can we move these conversions to `uint`? Is there any technical obstacle to do that?

primitive-types/src/lib.rs
```diff
@@ -0,0 +1,172 @@
...
+#[cfg(feature = "impl-rlp")] impl_fixed_hash_rlp!(H160, 20);
+
+#[cfg(feature = "std")]
+impl From<u64> for H160 {
+	fn from(val: u64) -> Self {
+		Self::from_low_u64_be(val)
+	}
+}
+
+impl From<H160> for H256 {
```
dvdplm: This should use `impl_fixed_hash_conversions!(H256, H160)` I think.

primitive-types/src/lib.rs
```diff
@@ -0,0 +1,172 @@
...
+construct_fixed_hash! {
+	/// Fixed-size uninterpreted hash type with 20 bytes (160 bits) size.
+	pub struct H160(20);
+}
+#[cfg(feature = "impl-serde")] impl_fixed_hash_serde!(H160, 20);
+#[cfg(feature = "impl-codec")] impl_fixed_hash_codec!(H160, 20);
+#[cfg(feature = "impl-rlp")] impl_fixed_hash_rlp!(H160, 20);
+
+#[cfg(feature = "std")]
+impl From<u64> for H160 {
```
dvdplm: @Robbepop maybe we should have this in `uint`?

primitive-types/src/lib.rs
```diff
@@ -0,0 +1,172 @@
...
+		ret.0[12..32].copy_from_slice(value.as_bytes());
+		ret
+	}
+}
+
+construct_fixed_hash! {
+	/// Fixed-size uninterpreted hash type with 32 bytes (256 bits) size.
+	pub struct H256(32);
+}
+#[cfg(feature = "impl-serde")] impl_fixed_hash_serde!(H256, 32);
```
dvdplm: style nit: I prefer these on two lines.

primitive-types/src/lib.rs
```diff
@@ -0,0 +1,172 @@
...
+#[cfg(feature = "impl-codec")] impl_fixed_hash_codec!(H256, 32);
+#[cfg(feature = "impl-rlp")] impl_fixed_hash_rlp!(H256, 32);
+
+#[cfg(feature = "std")]
```
dvdplm: Why does this need `std`?

primitive-types/src/lib.rs
```diff
@@ -0,0 +1,172 @@
...
+	fn from(val: u64) -> Self {
+		Self::from_low_u64_be(val)
+	}
+}
+
+construct_fixed_hash! {
+	/// Fixed-size uninterpreted hash type with 64 bytes (512 bits) size.
+	pub struct H512(64);
+}
+#[cfg(feature = "impl-serde")] impl_fixed_hash_serde!(H512, 64);
```
dvdplm: style nit: I prefer these on two lines.

primitive-types/src/lib.rs
sorpaas: I don't think we can -- it involves two uint types and in `uint` crate we can only deal with one that's defined.

primitive-types/src/lib.rs
Robbepop: This is problematic since in `uint` you can also define `uint` that are smaller than `u64` and thus making this conversion failing. However, `From` impls are guaranteed to never fail. So we would need to make this conversion lossy in order to guarantee that it never fails but I am not sure if this is explicitly wanted behaviour since this may introduce unwanted semantics.

primitive-types/src/lib.rs
Robbepop: I think this is because of `Self::from_low_u64_be` which itself requires `fixed-hash/byteorder` support. As far as I can see this crate tries to simplify crate features so that there is only something like `std` left which is enabled by default. I can understand this decision, however, one should keep in mind that this makes it impossible to use this API without std.

impls/codec/src/lib.rs
```diff
@@ -0,0 +1,52 @@
+// Copyright 2015-2017 Parity Technologies
```
bkchr: All the copyright years should be `2015-2018`.

impls/serde/src/serialize.rs
```diff
@@ -0,0 +1,120 @@
...
+			}
+
+			bytes.map_err(|e| E::custom(format_err(e)))
+		}
+
+		fn visit_string<E: de::Error>(self, v: String) -> Result<Self::Value, E> {
+			self.visit_str(&v)
+		}
+	}
+	// TODO [ToDr] Use raw bytes if we switch to RLP / binencoding
```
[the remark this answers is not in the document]
sorpaas: Let me try. This file was from the substrate codebase (and it was adopted from a file in `paritytech/primitives`, which I think was originally in `paritytech/parity`..)

Cargo.toml
```diff
@@ -17,5 +17,9 @@ members = [
 	"trace-time",
 	"trie-standardmap",
 	"triehash",
-	"uint"
+	"uint",
+	"impls/codec",
```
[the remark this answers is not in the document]
sorpaas: I moved them to `primitive-types/impls`. `rlp`, `codec` is about *encoding* while `serde` is about *serialization*, so I think maybe just call it `impls` there.

impls/serde/src/serialize.rs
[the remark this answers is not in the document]
sorpaas: I gave up... It just span across two different repositories for a really long time with file path changes, which makes it really difficult.

primitive-types/src/lib.rs
sorpaas: The issue is that `full_mul` involves two uint types. But in `uint` crate we can only construct one at a time and have zero information about the other.

primitive-types/src/lib.rs
sorpaas: We don't have that macro now and I don't think it's a good idea -- converting `H160` to `H256` is a special operation (that has agreement on Substrate and Parity Ethereum). We cannot apply it directly to any other types (otherwise, we can't answer issues like why is it zero-padded on the front? why not on the end or repeat the bytes?).

primitive-types/src/lib.rs
sorpaas: This from impl is fairly short so I think it's okay to just be there.

primitive-types/src/lib.rs
sorpaas: Fixed. It shouldn't be `std`.

primitive-types/src/lib.rs
sorpaas: Hmm I'm thinking -- maybe we should remove those `From<u64>` conversions? AFAIK it's only used in Substrate, but the issue with it is that it assumes big-endian encoding. To fix those in Substrate is trivial -- we just replace any `From<u64>` by the explicit function `from_low_u64_be`.

primitive-types/src/lib.rs
sorpaas: And see https://github.com/paritytech/parity-common/pull/86#discussion_r238522773. I think we should remove `From<u64>` defs.

impls/codec/src/lib.rs
sorpaas: I changed all touched files in this PR to be `2015-2018`, but left everything else unchanged, to avoid making this PR too big, and also remembered this was the guideline for changing copyright years.

primitive-types/src/lib.rs
dvdplm: > replace any `From<u64>` by the explicit function `from_low_u64_b`

I like this solution. Let's do that. :)

primitive-types/src/lib.rs
dvdplm: I meant [this](https://github.com/paritytech/parity-common/blob/59d9672df718cf42316b4445c641aa977461c028/fixed-hash/src/hash.rs#L817); it does a static assert on the type size and zero-pads in front. I agree this is arbitrary and not something we invented recently, but it is there and it's been used for a long time so I'd argue it's the "standard by convention" in our codebase. As long as we document the behaviour I can live with it.

The reason I'm being difficult here and elsewhere is that I'm taking the POV of a dev that is unfamiliar with the code structure and approaches the code base for the first time: I'm afraid they will grow frustrated with how code is spread out in different crates with a hard-to-grasp logic as to what stuff is found where. If at all possible I'd like logic to be in `uint`/`fixed-hash`.

sorpaas: Ah right. Fixed!

Cargo.toml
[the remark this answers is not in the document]
dvdplm: I still think "impls" is not a great name, but I won't make a fuss. :)

PR #87, opened 2018-12-03
Allow private uint definition and crunchy convenience fix

sorpaas:
* Use macro reexport feature so that crates that depends on uint doesn't need to declare crunchy dep any more.
* Change construct_uint definition to match `fixed-hash` so that users can add docs, visibility, and additional attributes.
* Remove `common` feature from default -- it's not usually used.

### DIFF
--- a/uint/benches/bigint.rs
+++ b/uint/benches/bigint.rs
@@ -16,9 +16,23 @@
 
 extern crate core;
 extern crate test;
+#[macro_use]
 extern crate uint;
 
-use uint::{U256, U512};
+construct_uint! {
+	pub struct U256(4);
+}
+
+construct_uint! {
+	pub struct U512(8);
+}
+
+impl U256 {
+	#[inline(always)]
+	pub fn full_mul(self, other: U256) -> U512 {
+		U512(uint_full_mul_reg!(U256, 4, self, other))
+	}
+}
 
 use test::{Bencher, black_box};
 
--- a/uint/Cargo.toml
+++ b/uint/Cargo.toml
@@ -18,11 +18,10 @@ crunchy = { version = "0.2", default-features = true }
 [features]
 default = ["std"]
 std = ["byteorder/std", "rustc-hex/std", "crunchy/std"]
-common = []
 
 [[example]]
 name = "modular"
 
 [[test]]
 name = "uint_tests"
-required-features = ["std", "common"]
+required-features = ["std"]
--- a/uint/Cargo.toml
+++ b/uint/Cargo.toml
@@ -16,7 +16,7 @@ quickcheck = { version = "0.6", optional = true }
 crunchy = { version = "0.2", default-features = true }
 
 [features]
-default = ["std", "common"]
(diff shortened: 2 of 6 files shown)


dvdplm:
>     * it's not usually used

Do we know which crate(s) does use it? Can you open tickets on the projects where it is used so we don't lose track of it when upgrading?

sorpaas:
@dvdplm No. I don't know any crate ever use this, and at least none of our projects use this feature.

dvdplm:
@Robbepop Do you know of any code that uses the `common` feature? Can we remove it you think?

dvdplm:
Only outstanding question is if we can remove the `common` feature or not. I think yes, but if @Robbepop could chime in on that too it'd be great. :)

dvdplm:
@sorpaas wrt to the `common` feature, what I am suggest is we remove it completely so that `uint` doesn't export any type at all. Hopefully there isn't any code out there that uses `uint::U256`; I think we only use `ethereum_types::U256` (or constructs the type itself, like substrate does) and the end-goal here is to switch all code to use `primitive_types::U256` anyway right?

If you prefer to do it step by step we can remove `common` from the default feature set and add a deprecation in `0.6` and then remove it completely in`0.7`. My advice is to remove it now though. 
What do you think?

uint/benches/bigint.rs
```diff
@@ -16,9 +16,23 @@
...
+	pub struct U256(4);
+}
+
+construct_uint! {
+	pub struct U512(8);
+}
+
+impl U256 {
+	#[inline(always)]
+	pub fn full_mul(self, other: U256) -> U512 {
```
sorpaas: It's possible to avoid this duplication once we have `primitive-types`. For now, I would rather keep it like this to avoid "circular PR".

Robbepop:
> @Robbepop Do you know of any code that uses the `common` feature? Can we remove it you think?

I don't know of any. It was introduced as default crate feature to stay backwards compatible but we may really consider to remove it in the end.

We could

1. Deprecate it in 0.6 (this) and remove it in 0.7 (next) version.
2. Remove it in 0.6 (this) since it is another minor version (but major changes are allowed since it is 0.x).

uint/benches/bigint.rs
Robbepop: good idea!

dvdplm: @sorpaas can you please add an issue for this so we don't forget? "Once PR #xx is merged, remove|deprecate "full_mul"… etc"?

dvdplm:
Tests fail. Otherwise lgtm.

uint/benches/bigint.rs
sorpaas: Added as: https://github.com/paritytech/parity-common/issues/91

larsrh:
`common` is used here: https://github.com/paritytech/pwasm-std/blob/335f9056d4e3be2a5ed1bf9710931296ae6a51ac/src/types.rs

PR #89, opened 2018-12-10
Initialize a crate for memory related tools

cheme:
This PR is a first step for heapsize removal/replacement https://github.com/paritytech/parity-ethereum/issues/9953 . It uses a fork of `malloc_size_of` (unpublished crate from servo project).

This PR create a new crate `parity-util-mem`:
- allows defining a global allocator (target should be jemalloc), by using a feature. (not that there is no priority so if multiple allocator feature are defined the compilation will panic in a global allocator redefinition error).
- plugs selected allocator to their heapsize measurement (selection of allocator with a feature).
- allows measurement alternative with only size_of or size_of_val : this can be activated quickly by setting `estimate-heapsize` feature.
- move code from parity-ethereum `mem` crate (`MemZero`), and and `clear_on_drop` reexport. This is from a previous split commit, I can put if out of this pr if needed.

After this crate cration, the parity-ethereum base code will basically look like: https://github.com/paritytech/parity-ethereum/compare/master...cheme:rem_heapsize . I may overuse the derive macro but I think it makes it more consistent (you do not need to check heapsize implementation when adding a field), there is also still some room for some implementations (especially if we want to use ConditionalMallocSizeOf, I need to test if there is some need).

Parity-common kvdb crate (otherwhise it will segfault for other allocator than malloc, because it still uses old heapsize directly).

The `malloc_size_of` usage is a bit awkward:
We copy code from servo project and patch it to remove some contents. We also switch `servoarc` to std arc.
I put a .sh file as documentation of the process and a .patch file to allow update from servo repo (I do not think there will be a lot).

The reason for this choice are :
- using as an external crate does not allow defining trait for custom types (parity-types) in the crate. That is very problematic when it comes to update things (even with local projects the dependency on parity-types are really difficult to change. Therefore including the trait in the parity crate let us implement the trait for those types and have an easier switching (we can consider moving the implementation in `fixed-hash` and `uint` at some point but it is difficult as a first step (if we plan to I shall feature gate those implementations to make the switch possible without republishing the crate).
- using as an external crate by specifying github address is also awkward due to the size of the servo github project (for `malloc_sizeof_derive` crate it is doable right now).
- the crate itself is really small.


mem/src/malloc_size.rs
```diff
@@ -0,0 +1,1151 @@
...
+//!   trait.
+//! - If an `Rc` or `Arc` should be measured only if it hasn't been seen
+//!   before, it should be measured via the `MallocConditionalSizeOf` trait.
+//! - Using universal function call syntax is a good idea when measuring boxed
+//!   fields in structs, because it makes it clear that the Box is being
+//!   measured as well as the thing it points to. E.g.
+//!   `<Box<_> as MallocSizeOf>::size_of(field, ops)`.
+
+
+// This file is patched at commit 5bdea7dc1c80790a852a3fb03edfb2b8fbd403dc DO NOT EDIT.
```
dvdplm: As I mentioned above, I think we should considering either forking or write our own crate inspired by this one. It seems like 90% of the code here is irrelevant to our needs and that many types we might want to measure need specialized code anyway.

Cargo.toml
```diff
@@ -17,5 +17,6 @@ members = [
 	"trace-time",
 	"trie-standardmap",
 	"triehash",
-	"uint"
+	"uint",
+  "mem",
```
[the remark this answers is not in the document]
cheme: will use the package name 'parity-util-mem'

mem/get_malloc_size_src.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+
+# script/process to update code from servo project (malloc_size_of)
+# untested, note that we do not use submodule due to size of git repo
+git clone https://github.com/servo/servo.git
```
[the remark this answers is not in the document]
cheme: Let's say the script is more a doc than a script, maybe I shall move it to the Readme.
Also if you know how to do a shallow clone up to a define commit reference, I will be interested.

mem/get_malloc_size_src.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+
+# script/process to update code from servo project (malloc_size_of)
```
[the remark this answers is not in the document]
cheme: It is somehow what I did, I just keep the delta to be able to synch simply, but in the long term it may prove useless and importing code somehow will be better.

mem/src/allocators.rs
```diff
@@ -0,0 +1,239 @@
...
+pub trait MallocSizeOfExt: MallocSizeOf {
+	fn m_size_of(&self) -> usize {
+		let mut ops = new_malloc_size_ops();
+		<Self as MallocSizeOf>::size_of(self, &mut ops)
+	}
+}
+
+impl<T: MallocSizeOf> MallocSizeOfExt for T {}
+
+/// we currently do not have use case where a conditional fn is use so
```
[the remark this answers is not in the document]
cheme: will remove in next commit (I got test to run but need first to use the trait in parity-ethereum and try to see the impact afterwards).

mem/src/allocators.rs
```diff
@@ -0,0 +1,239 @@
...
+		panic!("Running estimation this code should never be reached")
+	}
+
+	#[inline]
+	pub fn new_enclosing_size_fn() -> Option<VoidPtrToSizeFn> {
+		None
+	}
+}
+
+/// Get a new instance of a MallocSizeOfOps
```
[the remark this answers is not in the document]
cheme: I understood it as operators (not sure), but "allocation measurement operation" could make sense to.
In fact when trying to do conditional metering it does contain a state in the form of an `FnMut`, so it is probably not the best chosen name.

mem/src/impls.rs
```diff
@@ -0,0 +1,78 @@
...
+malloc_size_of_is_0!(std::time::Duration);
+
+malloc_size_of_is_0!(
+	U64, U128, U256, U512, H32, H64,
+	H128, H160, H256, H264, H512, H520,
+	Bloom
+);
+
+macro_rules! impl_elastic_array {
+	($name: ident, $dummy: ident, $size: expr) => (
```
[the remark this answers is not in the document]
cheme: It is used to copy paste the macro calls6 from elastic_array crate, I can change it.

mem/src/malloc_size.rs
```diff
@@ -0,0 +1,1151 @@
+// Copyright 2016-2017 The Servo Project Developers. See the COPYRIGHT
+// file at the top-level directory of this distribution and at
+// http://rust-lang.org/COPYRIGHT.
+//
+// Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
```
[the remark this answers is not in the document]
cheme: It seems like MIT is compatible and apache v2 is only compatible with GPLv3 (not gpl v2 but we are v3). So it looks fine.

mem/src/malloc_size.rs
cheme: - forking: doable but we need to get src/impl.rs and src/sizeof.rs in the fork, I prefer using a single crate (in the fork scenario I would consider reexported through parity-util-mem). The thing that I do not like with the fork is that we fork servo repo (thinking twice after trimming all the other stuff it is the same as a smaller repo, just with a huge history : I do not know if cargo is doing shallow clone).
- write our own crate: I am not comfortable with writing existing code, but this PR is in someway doing that.

tomaka:
> The reason for this choice are :

Why not copy-paste Servo's crate (with the proper licensing of course), instead of cloning it and patching it? It would be the first crate we fork.

cheme:
The crate is actually copied, I just wanted to keep it possible to synch some remove change easilly (but it may really not be necessary). So I guess it is possible (just need to remove the script and a comment in the copied code).

Proper way to do it should be to fork the crate, but it seems a bit too much when you consider the servo crate size.

tomaka:
Shouldn't we remove everything specific to Servo, though?

I was most in favour of not using a trait and having direct methods on everything, but the approach of this PR may be better. I'm in favour of it except for the fact that the crate is very large.

parity-util-mem/README.md
```diff
@@ -0,0 +1,16 @@
+# parity-util-mem
+
+Collection of memory related utilities.
+
+## Features
+
+- volatile-erase : Not set by default, `Memzero` struct will be erasing memory throug a simple [`write_volatile`](https://doc.rust-lang.org/std/ptr/fn.write_volatile.html) call.
```
pepyakin:
```suggestion
- volatile-erase : Not set by default, `Memzero` struct will be erasing memory through a simple [`write_volatile`](https://doc.rust-lang.org/std/ptr/fn.write_volatile.html) call.
```

parity-util-mem/malloc_size_of_derive/LICENSE-APACHE
```diff
@@ -0,0 +1,201 @@
...
+   To apply the Apache License to your work, attach the following
+   boilerplate notice, with the fields enclosed by brackets "[]"
+   replaced with your own identifying information. (Don't include
+   the brackets!)  The text should be enclosed in the appropriate
+   comment syntax for the file format. We also recommend that a
+   file or class name and description of purpose be included on the
+   same "printed page" as the copyright notice for easier
+   identification within third-party archives.
+
+Copyright [yyyy] [name of copyright owner]
```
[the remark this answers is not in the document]
cheme: I don't know, it's a direct copy from servo and I do not really want to change it.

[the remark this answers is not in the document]
pepyakin: Oh sorry, now I see!

cheme:
@tomaka I did change the patch to remove all servo related code (before it was deactivated by a feature), it makes the copied code smaller. Maybe some of the servo trait will prove useless (I did not plan to use the conditional one directly (need some testing to see if relevant)) and could also be removed.

parity-util-mem/src/sizeof.rs
```diff
@@ -0,0 +1,62 @@
...
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Parity.  If not, see <http://www.gnu.org/licenses/>.
+
+//! Estimation for heapsize calculation. Usable to replace call to allocator method (for some
+//! allocators or simply because we just need a determinist indicator of cunsumption).
+
+#[cfg(feature = "serde_only")]
+extern crate serde_bytes;
```
ordian: Do we use that crate?

parity-util-mem/src/sizeof.rs
```diff
@@ -0,0 +1,62 @@
+// Copyright 2015-2018 Parity Technologies (UK) Ltd.
```
ordian: Could you update year to 2019 here and above?

parity-util-mem/src/sizeof.rs
```diff
@@ -0,0 +1,62 @@
...
+// Parity is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Parity.  If not, see <http://www.gnu.org/licenses/>.
+
+//! Estimation for heapsize calculation. Usable to replace call to allocator method (for some
+//! allocators or simply because we just need a determinist indicator of cunsumption).
```
ordian:
```suggestion
//! allocators or simply because we just need a deterministic consumption measurement).
```

parity-util-mem/src/malloc_size.rs
```diff
@@ -0,0 +1,605 @@
...
+#[cfg(not(feature = "std"))]
+use alloc::string::String;
+#[cfg(not(feature = "std"))]
+mod std {
+  pub use core::*;
+  pub use alloc::collections;
+}
+
+#[cfg(feature = "std")]
+use std::sync as servo_arc;
```
tomaka: This name is a bit weird.

ordian: I guess it was done to minimize the diff, but is indeed confusing to readers of this code.

cheme: yes but because the diff is not really readable anymore, I will change that.

parity-util-mem/src/sizeof.rs
ordian: Looks like it wasn't removed from `sizeof.rs`

cheme:
I just realize that publishing this to crates.io (requires for dependency such as 'memorydb' (the only one that really need an update), means that I need to publish 'malloc_size_of_derive' (by design .
I see 3 scenarii here:
- ask servo team to publish 'malloc_size_of_derive' : probably not suitable as 'malloc_size_of' is unpublished.
- rename the copy of crate in the pr to 'parity-util-mem-derive' and publish it as a parity crate, I really don't like that because there is really no change to the code.
- rewrite the proc macro (not super interesting) and publish it as a parity crate.

tomaka:
Since we will have our local version of `malloc_size_of`, it makes sense to me to have our local version of `malloc_size_of_derive` as well.
If Servo's trait diverges from us, the proc macro will likely diverge as well.

ordian:
The windows CI failure is unrelated (I wonder whether we shouldn't have switched from appveyor).

@cheme Could you open a PR to switch parity-common to this crate before publishing it? I guess that would be a breaking change, so versions need to be bumped.

PR #92, opened 2019-01-11
Expose U128, add u128 conversions.

tomusdrw:
I think long term it would be good to use `u128` as words instead of `u64`

### DIFF
--- a/uint/src/uint.rs
+++ b/uint/src/uint.rs
@@ -305,11 +305,6 @@ macro_rules! impl_mul_from {
 		}
 	}
 }
-#[macro_export]
-#[doc(hidden)]
-macro_rules! impl_if {
-	( $t:tt ) => ( $t )
-}
 
 #[macro_export]
 #[doc(hidden)]
--- a/uint/src/uint.rs
+++ b/uint/src/uint.rs
@@ -347,11 +347,11 @@ pub fn split_u128(a: u128) -> (u64, u64) {
 #[macro_export]
 macro_rules! construct_uint {
 	( $(#[$attr:meta])* $visibility:vis struct $name:ident (1); ) => {
-		construct_uint!{ @construct $($attr)* $visibility struct $name (1); }
+		construct_uint!{ @construct $(#[$attr])* $visibility struct $name (1); }
 	};
 
 	( $(#[$attr:meta])* $visibility:vis struct $name:ident ( $n_words:tt ); ) => {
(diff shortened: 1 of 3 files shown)


uint/src/uint.rs
```diff
@@ -414,7 +464,7 @@ macro_rules! construct_uint {
 			pub fn as_u32(&self) -> u32 {
 				let &$name(ref arr) = self;
 				if (arr[0] & (0xffffffffu64 << 32)) != 0 {
-					panic!("Integer overflow when casting U256")
+					panic!("Integer overflow when casting to u64")
```
niklasad1:
```suggestion
					panic!("Integer overflow when casting to u32")
```

uint/src/uint.rs
```diff
@@ -346,7 +346,58 @@ pub fn split_u128(a: u128) -> (u64, u64) {
...
+					let mut ret = [0; $n_words];
+					ret[0] = value as u64;
+					ret[1] = (value >> 64) as u64;
+					$name(ret)
+				}
+			}
+
+			impl $crate::core_::convert::From<i128> for $name {
+				fn from(value: i128) -> $name {
+					match value >= 0 {
```
niklasad1: this could be `match (n as u128) >> 127 == 0` but the assembly is identical according to https://rust.godbolt.org/z/5kXf8K, so better keep it way more readable

uint/src/uint.rs
```diff
@@ -841,7 +890,7 @@ macro_rules! construct_uint {
 		impl_map_from!($name, i32, i64);
 		impl_map_from!($name, isize, i64);
 
-		// Converts from big endian representation of U256
+		// Converts from big endian representation of $name
```
sorpaas: This is a comment. Does macro name resolve for comments?

sorpaas:
@niklasad1 I think we moved `uint` from primitives repo to here, and that one is removed.

sorpaas:
Looks like test fails because metadata macro is broken somehow: https://travis-ci.org/paritytech/parity-common/jobs/478340732#L807

niklasad1:
@sorpaas 

All right I see, makes sense with your pending PR in the `primitives repo´.

uint/src/uint.rs
tomusdrw: Good point, seems it doesn't https://github.com/rust-lang/rust/issues/37903

PR #95, opened 2019-01-15
Bump primitive-types and uint version

sorpaas:
Bump uint v0.6.1 and primitive-types v0.1.6 versions. Changeset does not modify existing interface so I think a patch release is okay.


sorpaas:
> Well, doesn't it add new functionality i.e, <uint::128 operations>, primitives::U128?

Yeah, but I think the new features are small enough that we can just make a patch.

PR #97, opened 2019-01-16
Move `ethereum-types` to `parity-common` repo and more conversion functions for U128

sorpaas:
This completely moves `primitives` repo into this.

Now that `U128` is added to `primitive-types`, we can move those conversion functions for them from `ethereum-types`.


primitive-types/src/lib.rs
```diff
@@ -121,6 +121,19 @@ impl From<U256> for U512 {
 	}
 }
 
+impl From<U256> for U128 {
```
tomusdrw: I'm not a fan of panicking `From` conversions, it doesn't make any sense for me to have them in a general-purpose library like this.

primitive-types/src/lib.rs
```diff
@@ -136,6 +149,39 @@ impl From<U512> for U256 {
...
+	fn from(value: U128) -> U512 {
+		let U128(ref arr) = value;
+		let mut ret = [0; 8];
+		ret[0] = arr[0];
+		ret[1] = arr[1];
+		U512(ret)
+	}
+}
+
+impl From<U128> for U256 {
```
tomusdrw: For other types it seems that we also have conversions from references (see below), should we just implement for every `T: AsRef<U128>` or sth?

sorpaas:
> I don't really fully understand why paritytech/primitives still exists. Shouldn't we just put everything here or sth?

We can, and I think that would be a good idea. I create a PR that migrates `ethereum-types` with `primitive-types`.

primitive-types/src/lib.rs
sorpaas: Sorry replying late on this. The issue is that we currently already have `panic!` in a conversion that is actually needed `From<U512> for U256` in `evm`. Maybe we can keep this for now, and change everything to `TryFrom` once that's stabilized?

sorpaas:
I'm going to use this branch for migrating `ethereum-types` to `parity-common`, because the type conversion is dependent.

dvdplm:
@tomusdrw the rationale for parity-common is that it contains crates used in more than one parity project (e.g. parity-ethereum and substrate). Originally this was not the case for the crates in `primitives`. Maybe that is changing? If yes, then by all means we should move the crates here. If not, I'm much less convinced!

@sorpaas to make this easier to review, do you think you can point out the relevant commits for the `U128` work? Thanks!

primitive-types/src/lib.rs
tomusdrw: Why not have our own `TryFrom` trait specfic for this crate and migrate to the std one when it's stabilises?
What I mean is that here we introduce new code that we know is wrong and we say that we are going to fix it in the unspecficied future. I'm just a little bit afraid that it might bite us in the ass instead.

sorpaas:
@dvdplm https://github.com/paritytech/parity-common/pull/97/commits/53f3ec9af4c9a00cb8ec5a35e6bfa4ffe7cd4908

primitive-types/src/lib.rs
sorpaas: Added `TryFrom` trait and removed `panic!`

primitive-types/src/lib.rs
sorpaas: I haven't found a way to make it work -- I get conflicting impl if I try to put T inside `From`. I think it's only possible to implement one `From<T>` for a specific type, regardless of the where clause.

primitive-types/src/error.rs
```diff
@@ -0,0 +1,65 @@
...
+}
+
+/// Attempt to construct `Self` via a conversion.
+pub trait TryFrom<T>: Sized {
+    /// The type returned in the event of a conversion error.
+    type Error;
+
+    /// Performs the conversion.
+    fn try_from(value: T) -> Result<Self, Self::Error>;
+}
```
niklasad1: `TryInto/TryFrom` is likely to be in included in Rust 1.33 or 1.34 when this PR is merged create a ticket to remove these traits when they are included in Rust.

https://github.com/rust-lang/rust/issues/33417

ethbloom/Cargo.toml
```diff
@@ -0,0 +1,25 @@
+[package]
+name = "ethbloom"
+version = "0.6.0"
+authors = ["Parity Technologies <admin@parity.io>"]
+description = "Ethereum bloom filter"
+license = "MIT"
+documentation = "https://docs.rs/ethbloom"
+homepage = "https://github.com/paritytech/primitives"
+repository = "https://github.com/paritytech/primitives"
+
```
niklasad1: `homepage`  and `repository` needs to be updated to https://github.com/paritytech/parity-common

ethereum-types/Cargo.toml
```diff
@@ -0,0 +1,24 @@
+[package]
+name = "ethereum-types"
+version = "0.5.0"
+authors = ["Parity Technologies <admin@parity.io>"]
+license = "MIT"
+homepage = "https://github.com/paritytech/primitives"
```
niklasad1: Same here should be https://github.com/paritytech/parity-common

primitive-types/src/error.rs
sorpaas: Created https://github.com/paritytech/parity-common/issues/103

sorpaas:
The windows build always erred, but it looks like a CI issue.

PR #98, opened 2019-01-23
update parity-codec for primitive types

thiolliere:
this is a breaking change as parity-codec is exposed


thiolliere:
OK the CI failed two times on windows for too long without output but finally succeded.

Need one more review. Then I'll ask for a new release

bkchr:
@thiolliere yeah that is normal. Windows is still experimental in Travis.

PR #107, opened 2019-02-07
fix(serde deser): fix indexing into `str` bug

niklasad1:
[str](https://doc.rust-lang.org/std/primitive.str.html) consist of unicode characters which can be between 1-4 bytes
long and indexing assuming that all are 1 byte may panic

### DIFF
--- a/primitive-types/impls/serde/src/serialize.rs
+++ b/primitive-types/impls/serde/src/serialize.rs
@@ -81,7 +81,7 @@ pub fn deserialize_check_len<'de, D>(deserializer: D, len: ExpectedLen) -> Resul
 		}
 
 		fn visit_str<E: de::Error>(self, v: &str) -> Result<Self::Value, E> {
-			if v.len() < 2 || &v[0..2] != "0x" {
+			if v.len() < 2 || !v.starts_with("0x") {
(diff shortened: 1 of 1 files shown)


primitive-types/impls/serde/src/serialize.rs
```diff
@@ -81,7 +81,7 @@ pub fn deserialize_check_len<'de, D>(deserializer: D, len: ExpectedLen) -> Resul
 		}
 
 		fn visit_str<E: de::Error>(self, v: &str) -> Result<Self::Value, E> {
-			if v.len() < 2 || &v[0..2] != "0x" {
+			if v.len() < 2 || !v.starts_with("0x") {
```
niklasad1: This could be replaced `v.as_bytes()[0..2] != ['0' as u8, 'x' as u8]` if `starts_with` is too slow but I think it should be fast enough for such simple pattern

cheme: Shouldn't we just  : 
```suggestion
			if !v.starts_with("0x") {
```

niklasad1:
Ok to `merge` Windows build seems unrelated to this PR?

Issue #109, opened 2019-02-20
fixed-hash: cannot define Hash struct, like 'H256', in constants directly


koushiro:
fixed-hash v0.3.0 cannot define Hash struct in constants directly, like:

```rust
pub const TEST_HASH: H256 = H256([0xc5, 0xd2, 0x46, 0x01, 0x86, 0xf7, 0x23, 0x3c, 0x92, 0x7e, 0x7d, 0xb2, 0xdc, 0xc7, 0x03, 0xc0, 0xe5, 0x00, 0xb6, 0x53, 0xca, 0x82, 0x27, 0x3b, 0x7b, 0xfa, 0xd8, 0x04, 0x5d, 0x85, 0xa4, 0x70]);
```

In version 0.2.5, the definition of Hash structures is
```rust
pub struct $from (pub [u8; $size]);
```
but in version 0.3.0, the definition of Hash structure is 
```rust
$visibility struct $name ([u8; $n_bytes]);
```

Why does the new version of `fixed-hash` change the public field in the Hash structure to the private one?

Robbepop:
It was changed to have a better encapsulated state so that immutable references to hashes cannot change them. Also with Rust's progress on `const_fn` we can expect this to work in const context in the future. For now it doesn't. I myself am open to revert this again for the sake of usability.

koushiro:
@Robbepop Thanks for your reply! Maybe I can submit a PR to revert this and upgrade `fixed-hash` to v0.3.1?

PR #110, opened 2019-02-21
Fix fixed-hash

koushiro:

Fix the problem that the current version of  `fixed-hash` cannot be defined in constant directly.

related issue link: #109

### DIFF
--- a/fixed-hash/Cargo.toml
+++ b/fixed-hash/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "fixed-hash"
-version = "0.3.0"
+version = "0.3.1"
 authors = ["Parity Technologies <admin@parity.io>"]
 license = "MIT"
 homepage = "https://github.com/paritytech/parity-common"
--- a/fixed-hash/src/hash.rs
+++ b/fixed-hash/src/hash.rs
@@ -67,7 +67,7 @@ macro_rules! construct_fixed_hash {
 
 		impl<'a> From<&'a [u8; $n_bytes]> for $name {
 			/// Constructs a hash type from the given reference
-			/// to the bytes array of fixed length. 
+			/// to the bytes array of fixed length.
 			///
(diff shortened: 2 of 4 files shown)


koushiro:
[clabot:check]

Robbepop:
> Very much in favour of this. Without this change, the [`storage_keys!` macro](https://github.com/paritytech/pwasm-repo-contract/blob/master/contract/src/lib.rs) breaks.

Glad that you pointed this out!
However, on the other side the fix for `storage_keys` seems to be simple.

larsrh:
The way I'm working around this right now looks as follows:

```rust
pub struct Key {
    pub key: [u8; 32]
}

impl Key {
    pub fn new(key: [u8; 32]) -> Key {
        Key { key: key }
    }

    pub fn h256(&self) -> H256 {
        H256::from(self.key)
    }
}

#[macro_export]
macro_rules! storage_keys {
	() => {};
	($($name:ident),*) => {
		storage_keys!(0u8, $($name),*);
	};
	($count:expr, $name:ident) => {
		static $name: Key = Key { key: [$count, 0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0] };
	};
	($count:expr, $name:ident, $($tail:ident),*) => {
		static $name: Key = Key { key: [$count, 0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0] };
		storage_keys!($count + 1u8, $($tail),*);
	};
}
```

Essentially, I'm building my own key type for storage.

Robbepop:
I ll merge it as soon as CI is okay with it.

bkchr:
@Robbepop travis windows is always broken and after they kicked out all people, it will probably never work.

PR #114, opened 2019-03-01
tx-pool: replace error-chain with vanilla Error impl

ascjones:
Because of `Error::cause` deprecation in 1.33. 

Re: https://github.com/paritytech/parity-ethereum/issues/10302, https://github.com/paritytech/substrate/issues/1547

### DIFF
--- a/transaction-pool/src/error.rs
+++ b/transaction-pool/src/error.rs
@@ -14,15 +14,11 @@
 // You should have received a copy of the GNU General Public License
 // along with Parity.  If not, see <http://www.gnu.org/licenses/>.
 
-/// TODO: Consider making Hash generic for Error - legacy of error-chain
-/// So the hashes are converted to debug strings for easy display.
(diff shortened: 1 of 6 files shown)


transaction-pool/src/error.rs
```diff
@@ -14,34 +14,41 @@
...
+
+/// Transaction Pool Error
+#[derive(Debug)]
+pub enum Error<Hash: fmt::Debug + fmt::LowerHex> {
+	/// Transaction is already imported
+	AlreadyImported(Hash),
+	/// Transaction is too cheap to enter the queue
+	TooCheapToEnter(Hash, String),
+	/// Transaction is too cheap to replace existing transaction that occupies the same slot.
+	TooCheapToReplace (Hash, Hash),
```
bkchr:
```suggestion
	TooCheapToReplace(Hash, Hash),
```

PR #116, opened 2019-03-21
Extract `should_replace` from Scoring and supply pooled txs from same sender

ascjones:
`should_replace` does not use the computed score, and is actually called only on `import` (when pool limits are reached).

Further, we pass other pooled transactions from the respective senders of the `old` and `new` transactions which are being compared. This would allow an implementation to also compare the `Readiness` of the transactions when deciding if the `new` should push out the `old`.

### DIFF
--- a/transaction-pool/src/pool.rs
+++ b/transaction-pool/src/pool.rs
@@ -131,7 +131,9 @@ impl<T, S, L> Pool<T, S, L> where
 	/// NOTE: The transaction may push out some other transactions from the pool
 	/// either because of limits (see `Options`) or because `Scoring` decides that the transaction
 	/// replaces an existing transaction from that sender.
-	/// If any limit is reached the transaction with the lowest `Score` is evicted to make room.
+	///
+	/// If any limit is reached the transaction with the lowest `Score` will be compared with the
+	/// new transaction via the supplied `ShouldReplace` implementation and may be evicted.
 	///
 	/// The `Listener` will be informed on any drops or rejections.
 	pub fn import(&mut self, transaction: T, replace: &mut ShouldReplace<T>) -> error::Result<Arc<T>, T::Hash> {
--- a/transaction-pool/src/pool.rs
+++ b/transaction-pool/src/pool.rs
@@ -22,7 +22,7 @@ use error;
 use listener::{Listener, NoopListener};
 use options::Options;
 use ready::{Ready, Readiness};
-use scoring::{self, Scoring, ScoreWithRef, ShouldReplace};
+use scoring::{self, Scoring, ScoreWithRef, ShouldReplace, ReplaceTransaction};
 use status::{LightStatus, Status};
 use transactions::{AddResult, Transactions};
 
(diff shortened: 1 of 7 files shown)


transaction-pool/src/replace.rs
```diff
@@ -0,0 +1,53 @@
...
+
+//! When queue limits are reached, decide whether to replace an existing transaction from the pool
+
+use pool::Transaction;
+use scoring::Choice;
+
+/// Encapsulates a transaction to be compared, along with pooled transactions from the same sender
+pub struct ReplaceTransaction<'a, T> {
+    /// The transaction to be compared for replacement
+    pub transaction: Transaction<T>,
```
tomusdrw: Can't we borrow here as well?

ascjones:
Bump to version `2.0`?

transaction-pool/src/pool.rs
```diff
@@ -283,22 +286,32 @@ impl<T, S, L> Pool<T, S, L> where
...
-					return Err(error::Error::TooCheapToEnter(transaction.hash().clone(), format!("{:#x}", old.score)))
-				},
+			Some(old) => {
+                let txs = &self.transactions;
+                let get_replace_tx = |tx| {
+                    let sender_txs = txs.get(transaction.sender()).map(|txs| txs.iter().as_slice());
+                    ReplaceTransaction::new(tx, sender_txs)
+                };
+                let old_replace = get_replace_tx(&old.transaction);
+                let new_replace = get_replace_tx(transaction);
```
tomusdrw: Seems that it uses spaces for indentation,could you reformat?

tomusdrw:
Seems like there is still an issue with tests:
```rust
error[E0053]: method `should_replace` has an incompatible type for trait
  --> transaction-pool/src/tests/helpers.rs:77:20
   |
77 |     fn should_replace(&mut self, old: &ReplaceTransaction<Transaction>, new: &ReplaceTransaction<Transaction>) -> scoring::Choice {
   |                       ^^^^^^^^^ types differ in mutability
   | 
  ::: transaction-pool/src/replace.rs:52:23
   |
52 |     fn should_replace(&self, old: &ReplaceTransaction<T>, new: &ReplaceTransaction<T>) -> Choice;
   |                       ----- type in trait
   |
   = note: expected type `fn(&tests::helpers::DummyScoring, &replace::ReplaceTransaction<'_, tests::Transaction>, &replace::ReplaceTransaction<'_, tests::Transaction>) -> scoring::Choice`
              found type `fn(&mut tests::helpers::DummyScoring, &replace::ReplaceTransaction<'_, tests::Transaction>, &replace::ReplaceTransaction<'_, tests::Transaction>) -> scoring::Choice`
help: consider change the type to match the mutability in trait
   |
77 |     fn should_replace(&self, old: &ReplaceTransaction<Transaction>, new: &ReplaceTransaction<Transaction>) -> scoring::Choice {
   |                       ^^^^^
```

tomusdrw:
Merging since appveyor failures are unrelated and travis only failed the windows build (due to timeout).

PR #127, opened 2019-04-24
Use newly stablized TryFrom trait for primitive-types

sorpaas:
Remove custom `TryFrom`/`TryInto`/`Never` and use stable `TryFrom`.


primitive-types/src/lib.rs
```diff
@@ -14,9 +14,8 @@
 
 #![cfg_attr(not(feature = "std"), no_std)]
 
-mod error;
-
-pub use error::{Error, TryFrom, TryInto, Never};
+#![cfg(feature = "std")]
```
niklasad1: is this really correct? 
why enforce `feature = std` in the crate? 

I guess it should be `#[cfg(feature = "std")]`?

sorpaas: I tested in edition 2015 `#[cfg(feature = "std")]` is still necessary, otherwise compiler would complain `core` being reimported when building no_std.

PR #128, opened 2019-04-26
Add NumEntries trait to kvdb and impl it for rocksdb

dvdplm:
Currently there is no way of knowing how many keys there are in a DB. RocksDB does not have a means to retrieve the exact number of keys in a DB (would require a full compaction and key scan, see [this](https://github.com/facebook/rocksdb/wiki/RocksDB-FAQ)) but they do provide an "estimate". It is unclear to me just how much of an approximation this estimate is, but in simple tests it seems accurate (and regardless there is no better alternative).

This PR builds upon [this](https://github.com/paritytech/parity-common/pull/120) PR and adds a `NumEntries` trait to `kvdb` and implements it for RocksDB in the last commit. However, `parity-rocksdb` does not currently expose the necessary API calls to fetch the estimate (or any other runtime statistic) so this PR cannot be merged before [this](https://github.com/paritytech/rust-rocksdb/pull/31) is merged and published.


dvdplm:
Closing in favour of https://github.com/paritytech/parity-common/pull/285

PR #129, opened 2019-04-29
Benchmarks comparing LMDB and RocksDB

dvdplm:


kvdb-rocksdb/benches/kvdb-rocksdb.rs
```diff
@@ -0,0 +1,121 @@
...
+        ParameterizedBenchmark::new(
+            "payload size",
+            move |b, payload_size| {
+                let cfg = DatabaseConfig::with_columns(Some(1u32));
+                let db = Database::open(&cfg, path).unwrap();
+                let v = randbytes(*payload_size);
+                b.iter(move || {
+                    let mut batch = db.transaction();
+                    let k = randbytes(32); // All ethereum keys are 32 byte
+                    batch.put(None, &k, &v);
```
ordian: We may want to measure batch insert with more than one put

kvdb-lmdb/benches/kvdb-lmdb.rs
```diff
@@ -0,0 +1,125 @@
...
+    let path = "./benches/dbs/bench-lmdb-empty";
+    let path2 = path.clone();
+    c.bench(
+        "empty DB, 32 byte keys",
+        ParameterizedBenchmark::new(
+            "payload size",
+            move |b, payload_size| {
+                let cfg = DatabaseConfig::new(1u32);
+                let db = Database::open(&cfg, path).unwrap();
+                let v = randbytes(*payload_size);
```
ordian: Instead of using one value of `payload_size`, it would make sense to insert batch of `payload_size / 200` values of size 200 (and probably rename `payload_size` to `batch_size`).

kvdb-lmdb/benches/kvdb-lmdb.rs
```diff
@@ -0,0 +1,125 @@
...
+    // Try to avoid re-creating the test DB for every test run.
+    if !db_ok {
+        println!("Creating benchmark DB");
+        let (elapsed, _) = measure_time(|| {
+            let batches = size / batch_size;
+            for b in 0..batches {
+                println!("Batch: {}/{} – inserting indices {} –> {}", b, batches, b*batch_size, (b+1)*batch_size);
+                let mut tr = DBTransaction::with_capacity(batch_size);
+                for i in b*batch_size..(b+1)*batch_size {
+                    let key = b*batch_size + i;
```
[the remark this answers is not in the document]
dvdplm: No, this is ok I think. I want the DB to be filled with numeric keys in sequence (so they are sorted and deterministic and known before hand, with an eye to future read tests).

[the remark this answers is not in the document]
ordian: but with `i`, they will be in sequence 0..size, no?

kvdb-lmdb/benches/kvdb-lmdb.rs
dvdplm: Definitely need something better than a hard coded `200` here, but I suspect the payload size distribution is fairly important and I want to collect more data on it. So I'll return to this.

(I also think this "empty DB" test is fairly useless, will prolly remove it entirely)

kvdb-rocksdb/benches/kvdb-rocksdb.rs
dvdplm: Agreed, need to know more about how many ops we typically have in our batches though.

kvdb-lmdb/benches/kvdb-lmdb.rs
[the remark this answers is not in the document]
dvdplm: ofc, I'm dumb.

PR #155, opened 2019-05-15
[parity-crypto] Derive Debug for digests and signatures

dvdplm:
Based on #152, only https://github.com/paritytech/parity-common/pull/155/commits/20e8f69b092833f4720f1101827689c3bc563974 is new here.

### DIFF
--- a/parity-crypto/src/hmac/mod.rs
+++ b/parity-crypto/src/hmac/mod.rs
@@ -44,20 +44,19 @@ impl<T> Deref for Signature<T> {
 pub struct SigKey<T>(KeyInner, PhantomData<T>);
 
 enum KeyInner {
-	Sha256(rhmac::Hmac<rsha2::Sha256>),
-	Sha512(rhmac::Hmac<rsha2::Sha512>),
+	Sha256(Vec<u8>),
+	Sha512(Vec<u8>),
 }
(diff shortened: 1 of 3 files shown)


parity-crypto/src/hmac/mod.rs
```diff
@@ -44,19 +46,19 @@ impl<T> Deref for Signature<T> {
 pub struct SigKey<T>(KeyInner, PhantomData<T>);
 
 enum KeyInner {
-	Sha256(GenericArray<u8, U64>),
-	Sha512(GenericArray<u8, U128>),
+	Sha256(Vec<u8>),
```
niklasad1: Can you explain why `GenericArray<u8, T>` is replaced by `Vec<u8`?

dvdplm: Absolutely, should have been in the PR description. When replacing the HMAC code (used ring) with the `hmac` crate a change was introduced that required calling code to provide a full 128/64-bytes long key for SHA512/SHA256 (stored in the form of a `GenericArray`, sized to the digest block length). 
Using a `Vec` to store the incoming key removes the fixed size requirement and using `new_varkey()` ensures too long/too short keys are handled correctly.

PR #156, opened 2019-05-15
[parity-crypto] Add tests to hmac

dvdplm:
Add test vectors from IETF RFC4231


dvdplm:
#157 included everything here. Closing.

PR #161, opened 2019-05-17
Don't use local dep

dvdplm:


dvdplm:
Following further discussion with @ordian I've come around to his pov that it is not necessarily bad to have path dependencies. Closing.

PR #167, opened 2019-05-27
[rlp] Add no_std support

koushiro:

# What have I changed?

- Add `no_std` support for `rlp` crate
- Migrate `rlp` crate to 2018 edition
- Remove some test cases that depend on `primitive-types` (These test cases should be added to `primitive-types` instead of `rlp`)

### DIFF
--- a/rlp/Cargo.toml
+++ b/rlp/Cargo.toml
@@ -5,12 +5,15 @@ description = "Recursive-length prefix encoding, decoding, and compression"
 repository = "https://github.com/paritytech/parity-common"
 license = "MIT/Apache-2.0"
 authors = ["Parity Technologies <admin@parity.io>"]
+edition = "2018"
 
 [dependencies]
-byteorder = "1.0"
-rustc-hex = {version = "2.0", default-features = false }
+byteorder = { version = "1", default-features = false }
+rustc-hex = { version = "2.0", default-features = false }
 
-[dev-dependencies]
-hex-literal = "0.1"
-primitive-types = { path = "../primitive-types", version = "0.3", features = ["impl-rlp"] }
-
+[features]
+default = ["std"]
+std = [
+    "byteorder/std",
+    "rustc-hex/std",
+]
--- a/rlp/benches/rlp.rs
+++ b/rlp/benches/rlp.rs
@@ -14,11 +14,8 @@
 
 #![feature(test)]
 
-extern crate ethereum_types;
-extern crate rlp;
 extern crate test;
 
-use ethereum_types::U256;
 use rlp::{RlpStream, Rlp};
 use test::Bencher;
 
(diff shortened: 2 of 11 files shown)


rlp/Cargo.toml
```diff
@@ -5,12 +5,15 @@ description = "Recursive-length prefix encoding, decoding, and compression"
...
 license = "MIT/Apache-2.0"
 authors = ["Parity Technologies <admin@parity.io>"]
+edition = "2018"
 
 [dependencies]
-byteorder = "1.0"
-rustc-hex = {version = "2.0", default-features = false }
-
-[dev-dependencies]
-hex-literal = "0.1"
```
dvdplm: Why prefer `rustc-hex` to `hex-literal`?

rlp/benches/rlp.rs
```diff
@@ -106,13 +80,13 @@ fn bench_stream_1000_empty_lists(b: &mut Bencher) {
 fn bench_decode_1000_values(b: &mut Bencher) {
 	let mut stream = RlpStream::new_list(1000);
 	for _ in 0..1000 {
-		stream.append(&U256::from(1));
+		stream.append(&1u64);
```
dvdplm: This changes the amount of data tested, making it impossible to compare tests before/after this change.

In general I'm not sure why the dev dependencies can't depend on `primitive-types` (or any other code, be it `no_std` or not). I think the benchmark use the big uint types because it is such a common use case for the crate. What's the point of having the benchmarks be `no_std`?

rlp/src/rlpin.rs
```diff
@@ -387,11 +392,13 @@ impl<'a> BasicDecoder<'a> {
...
 mod tests {
-	use {Rlp, DecoderError};
+	#[cfg(not(feature = "std"))]
+	use alloc::format;
+	use super::*;
 
 	#[test]
 	fn test_rlp_display() {
-		let data = hex!("f84d0589010efbef67941f79b2a056e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421a0c5d2460186f7233c927e7db2dcc703c0e500b653ca82273b7bfad8045d85a470");
+		let data: Vec<u8> = rustc_hex::FromHex::from_hex("f84d0589010efbef67941f79b2a056e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421a0c5d2460186f7233c927e7db2dcc703c0e500b653ca82273b7bfad8045d85a470").unwrap();
```
dvdplm: As mentioned above I'm not sure why this is better?

rlp/Cargo.toml
koushiro: I think `rustc-hex` is enough to encode/decode hex. There is no need to introduce more dependencies, even if it is just `dev-dependencies`.

koushiro:
@dvdplm Thanks for reviewing my code!

I removed the `primitive-types` dependency because `H160` and `U256` are custom types in `primitive-types` crate. If we want to test their rlp codec, I think we should write tests in `primitive-types` instead of in `rlp` crate.

For example, if you define a structure that implements a `Serialize/Deserialize` trait and you want to know if the serialization and deserialization of the structure is as expected, you should write tests in the crate of the structure you defined, instead of submitting a PR to request the `serde-rs` maintainers to add a `dev-dependencies` to test your code.

koushiro:
I just make tests are `no_std` compatible, which is a good indication that the code is no problem in the `no_std` environment. As for the benchmarks of `U256` type, I think it should also be written in `primitive-types` crate for the same reason above.

rlp/Cargo.toml
dvdplm: Tests need to be as readable as possible, which is why I think`hex-literal` is better.

dvdplm: (The new `0.2` of `hex-literal` is edition compatible too!)

dvdplm:
@koushiro no, thank you for contributing! :)

So the context for my hesitation about moving the tests is that we've seen several instances of good PRs come in that look perfectly justified and that seem like an easy win and so they are merged but then when someone actually tries to use the new versions in production code, issues pop up. Most of the time it's ergonomic quirks but sometimes in the past we've merged code that turned out to be impossible to use higher up in the stack. I don't actually know what the best way to tackle this is – we can hardly expect all contributor to go through the hassle of upgrading depending crates/projects. The tests you remove here – and I totally understand your reasoning – makes me wonder if we're not removing one barrier to such unfortunate merges: the tests exercise a portion of the stack above itself so we know that part is sort of ok.

The obvious objection is that these are unit tests, not integration tests. And of course you and @sorpaas are right: these tests logically belong in `primitive-types`.

Maybe we can meet half-way:

1. revert the test removals here and keep only the `no_std` parts
1. in a separate PR, move the unit tests to where they belong and at the same time add a doc-test here that proves that `rlp` + `primitive-types` still work

Would that work for you?

koushiro:
@dvdplm @sorpaas I will add more tests (include removed tests in this PR) into `primitive-types` in another PR. PTAL again.

dvdplm:
> I will add more tests (include removed tests in this PR) into `primitive-types` in another PR

Let us know when this other PR is ready so we can unblock this one. :)

dvdplm:
@koushiro I'll close this for now. Feel free to re-open at any time with the last changes requested.

PR #210, opened 2019-09-02
Move ethkey crypto utils to parity crypto crate

grbIzl:
The intention of this PR is to remove dependency to accounts\ethkey for parity components, that use crypto utils from it. That would allow further decomposition of parity repo. Including move of secret store  into the separate repo.
Additional comments:
1. I straightforwardly merged code from ethkey into parity crypto, that is responsible for work with key pairs, encryption etc.

- I left wallets code (brain, brain prefix etc) in ethkey. Also left password.rs there
- I didn't do any renaming and reorganization of parity crypto crate. May be it's worth doing, but I don't have good ideas and would be grateful for any suggestions.

2. I'm not going to consider substrate code, related to the same crypto primitives. It's left for the further development.


parity-crypto/Cargo.toml
```diff
@@ -15,6 +15,9 @@ harness = false
 
 [dependencies]
 tiny-keccak = "1.4"
+eth-secp256k1 = { git = "https://github.com/paritytech/rust-secp256k1" }
+ethereum-types = "0.6.0"
```
ordian: do we want to use the latest version? (I know we're currently using 0.6 in parity-ethereum, but upgrading that shouldn't be hard)

parity-crypto/Cargo.toml
```diff
@@ -24,6 +27,9 @@ aes = "0.3.2"
 aes-ctr = "0.3.0"
 block-modes = "0.3.3"
 pbkdf2 = "0.3.0"
+quick-error = "1.2.2"
```
ordian: We've removed quick-error in #149, let's not reintroduce it again as it's deprecated.

parity-crypto/Cargo.toml
```diff
@@ -15,6 +15,9 @@ harness = false
 
 [dependencies]
 tiny-keccak = "1.4"
+eth-secp256k1 = { git = "https://github.com/paritytech/rust-secp256k1" }
```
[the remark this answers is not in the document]
ordian: Also it probably doesn't compile to WASM, so we want to hide behind a feature flag.

parity-crypto/Cargo.toml
```diff
@@ -24,6 +27,9 @@ aes = "0.3.2"
 aes-ctr = "0.3.0"
 block-modes = "0.3.3"
 pbkdf2 = "0.3.0"
+quick-error = "1.2.2"
+rand = "0.6"
```
ordian: Do we want `0.7` (we're making a breaking anyway)?

parity-crypto/src/secret.rs
```diff
@@ -0,0 +1,304 @@
...
+		if key.len() != 32 {
+			return None
+		}
+		let mut h = H256::zero();
+		h.as_bytes_mut().copy_from_slice(&key[0..32]);
+		Some(Secret { inner: h })
+	}
+
+	/// Creates zero key, which is invalid for crypto operations, but valid for math operation.
+	pub fn zero() -> Self {
```
dvdplm: Seems unwise to provide this function? It'd be great to refactor code using it (as we're making a breaking change anyway).

parity-crypto/src/secret.rs
```diff
@@ -0,0 +1,304 @@
...
+// (at your option) any later version.
+
+// Parity Ethereum is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Parity Ethereum.  If not, see <http://www.gnu.org/licenses/>.
+
```
dvdplm: Missing module docs. And name is too generic.

parity-crypto/src/secret.rs
```diff
@@ -0,0 +1,304 @@
...
+		Ok(secret.into())
+	}
+
+	/// Checks validity of this key.
+	pub fn check_validity(&self) -> Result<(), Error> {
+		self.to_secp256k1_secret().map(|_| ())
+	}
+
+	/// Inplace add one secret key to another (scalar + scalar)
+	pub fn add(&mut self, other: &Secret) -> Result<(), Error> {
```
dvdplm: There seems to be a lot of similar mathy methods here as well as in `math.rs`: maybe there's a trait here waiting to be discovered and impl'd for `Secret` and `Public`? `add`, `sub`, `mul`, `neg`, `is_valid` etc?

parity-crypto/src/signature.rs
```diff
@@ -0,0 +1,314 @@
...
+// (at your option) any later version.
+
+// Parity Ethereum is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Parity Ethereum.  If not, see <http://www.gnu.org/licenses/>.
+
```
dvdplm: Missing module docs and maybe a bit too generic name?

parity-crypto/src/signature.rs
```diff
@@ -0,0 +1,314 @@
...
+impl Hash for Signature {
+    fn hash<H: Hasher>(&self, state: &mut H) {
+    	H520::from(self.0).hash(state);
+    }
+}
+
+impl Clone for Signature {
+    fn clone(&self) -> Self {
+		Signature(self.0)
```
dvdplm: Odd looking whitespace. Shouldn't this be `Signature(self.0.clone())`?

ordian:
Also see #80 for previous attempt to add secp256k1 to parity-crypto (cc @cheme).

grbIzl:
> Also see #80 for previous attempt to add secp256k1 to parity-crypto (cc @cheme).

IMHO @cheme was too ambitious in his PR :-) I want to achieve simpler objectives (move common code from ethkey and don't break anything dependent on parity-crypto). And we need to have it done, if we want to develop secret store further as a separate project. So I hope to bring it to the (happy) end.

grbIzl:
> One thing I'm not clear about is if the code added here is entierly ethereum-specific or if there is stuff that is useful to other projects as well. You mentioned `secret-store` might be extracted after this. Are there other Parity projects that use/could use the crypto here?

The primary beneficiary of this PR is secret store indeed. I'm not aware of other projects for sure. At the same time, we wanted to extract account management, removing this dependency from it could help.

grbIzl:
@dvdplm thanks for comments! It will be good to cleanup this code, as it was not touched for a while.

dvdplm:
@grbIzl  I really appreciate the effort and the goal here! FWIW I tried something similar a while back and failed. 
It's a substantial chunk of code added though so I want to collect opinions from more people on where we want this to live. I think it's good to get as many eyes on it as possible so we avoid bitrot.

parity-crypto/Cargo.toml
cheme: I was wondering if it is doable to avoid 'ethereum-types' dependency. That would probably involve some slight function changes, and more importantly making 'message' a generic type. That is only an idea/question (may even be a bad idea due to possible misuse of message of unmatching len).

parity-crypto/Cargo.toml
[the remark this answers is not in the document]
cheme: Iirc, on #80 , there was some work to make https://crates.io/crates/libsecp256k1 'no_std' compatible (even for the secret store code) so I would be pretty confident in saying the switch is is possible.
Not sure it is needed for future 'secret-store' plan.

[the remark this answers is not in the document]
ordian: Before using `libsecp256k1` in parity-ethereum, we need to make sure it won't introduce a perf regression.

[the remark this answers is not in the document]
cheme: Last time it was checked, it does. It probably still does, (cc @sorpaas ), but I am not sure it is a significant change, and I do not know if performance is that critical for the secret store use case (for wasm32 it is a more straightforward choice).

dvdplm:
> Are there other Parity projects that use/could use the crypto here?

Substrate uses `parity-crypto` in `substrate-keystore`, @grbIzl, we need to make sure that when this PR is in good shape we submit upgrade PRs to both `parity-ethereum` and `substrate`.

parity-crypto/Cargo.toml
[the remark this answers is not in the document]
dvdplm: Ok, let's proceed without switching but I think this PR would be a good time to add some benchmarks so we can get a baseline understanding and avoid regressions.

parity-crypto/src/extended.rs
```diff
@@ -0,0 +1,514 @@
...
+pub enum Derivation<T: Label> {
+	/// Soft key derivation (allow proof of parent)
+	Soft(T),
+	/// Hard key derivation (does not allow proof of parent)
+	Hard(T),
+}
+
+impl From<u32> for Derivation<u32> {
+	fn from(index: u32) -> Self {
+		if index < (2 << 30) {
```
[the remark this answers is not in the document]
grbIzl: I'm also not a big fan of binary math, but here it makes sense, because of derivation definition. I've added more details about it here and for the whole module

parity-crypto/src/extended.rs
```diff
@@ -0,0 +1,514 @@
...
+
+		let serialized = new_public.serialize_vec(&SECP256K1, false);
+
+		Ok((
+			H512::from_slice(&serialized[1..65]),
+			new_chain_code,
+		))
+	}
+
+	fn sha3(slc: &[u8]) -> H256 {
```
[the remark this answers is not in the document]
grbIzl: It's used in the next method

parity-crypto/src/secret.rs
grbIzl: It's used as a default value for Secret (mostly in tests)

parity-crypto/src/secret.rs
grbIzl: But it's different algebraic groups (scalar and elliptic curve). I'm afraid, that general trait for them could lead to mistreatments

grbIzl:
I put newly added dependencies under the build feature (publickey).

The only not implemented major thing is including of ethereum_types crate. H256 is used heavily in Signature and Secret classes and I would need to copypaste its functionality there. So for now I leave it as a dependency (but put under the same feature)

parity-crypto/src/publickey/error.rs
```diff
@@ -0,0 +1,99 @@
...
+	}
+}
+
+impl Into<String> for Error {
+	fn into(self) -> String {
+		format!("{}", self)
+	}
+}
+
+impl From<::std::io::Error> for Error {
```
ordian: could you update all imports to 2018 edition for consistency? (meaning removing `::` prefix in this case)

parity-crypto/src/secret.rs
dvdplm: Mistreatments, can you elaborate on that? How could it be abused?

parity-crypto/src/publickey/ec_math_utils.rs
```diff
@@ -0,0 +1,131 @@
...
+		let public1 = Random.generate().unwrap().public().clone();
+		let public2 = Random.generate().unwrap().public().clone();
+
+		let mut sum = public1.clone();
+		public_add(&mut sum, &public2).unwrap();
+		public_sub(&mut sum, &public2).unwrap();
+
+		assert_eq!(sum, public1);
+	}
+}
```
dvdplm: There are a bunch of tests missing right? valid, multiplication, set_public etc?

parity-crypto/src/publickey/ecdh.rs
```diff
@@ -0,0 +1,37 @@
...
+
+// Parity Ethereum is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Parity Ethereum.  If not, see <http://www.gnu.org/licenses/>.
+
+//! ECDH scheme functions for key agreement
```
dvdplm:
```suggestion
//! ECDH key agreement scheme implemented as a free function.
```

parity-crypto/src/publickey/ecdh.rs
```diff
@@ -0,0 +1,37 @@
...
+		let mut temp = [4u8; 65];
+		(&mut temp[1..65]).copy_from_slice(&public[0..64]);
+		temp
+	};
+
+	let publ = key::PublicKey::from_slice(context, &pdata)?;
+	let sec = key::SecretKey::from_slice(context, secret.as_bytes())?;
+	let shared = ecdh::SharedSecret::new_raw(context, &publ, &sec);
+
+	Secret::from_unsafe_slice(&shared[0..32])
```
dvdplm: This has a scary name. Remind me, what is unsafe about it? (Maybe we need a comment here)

parity-crypto/src/publickey/ecdh.rs
```diff
@@ -0,0 +1,37 @@
...
+// You should have received a copy of the GNU General Public License
+// along with Parity Ethereum.  If not, see <http://www.gnu.org/licenses/>.
+
+//! ECDH scheme functions for key agreement
+
+use secp256k1::{self, ecdh, key};
+use super::{Error, Secret, Public, SECP256K1};
+
+/// Agree on a shared secret
+pub fn agree(secret: &Secret, public: &Public) -> Result<Secret, Error> {
```
dvdplm: I think this is a "hot" call. Would be great to have a benchmark for it.

parity-crypto/src/publickey/ecdsa_signature.rs
```diff
@@ -0,0 +1,316 @@
...
+
+// Parity Ethereum is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Parity Ethereum.  If not, see <http://www.gnu.org/licenses/>.
+
+//! ECDSA signature
```
dvdplm: Better docs pls.

parity-crypto/src/publickey/ecdsa_signature.rs
```diff
@@ -0,0 +1,316 @@
...
+	pub fn s(&self) -> &[u8] {
+		&self.0[32..64]
+	}
+
+	/// Get the recovery byte.
+	pub fn v(&self) -> u8 {
+		self.0[64]
+	}
+
+	/// Encode the signature into RSV array (V altered to be in "Electrum" notation).
```
dvdplm: "Electrum" as in the bitcoin wallet? Can you elaborate on what we use this for? And perhaps bake it into the comment here?

parity-crypto/src/publickey/ecdsa_signature.rs
```diff
@@ -0,0 +1,316 @@
...
+			return Signature::default();
+		}
+
+		let mut sig = [0u8; 65];
+		sig.copy_from_slice(data);
+		sig[64] -= 27;
+		Signature(sig)
+	}
+
+	/// Create a signature object from the sig.
```
dvdplm: "Create a signature object from the sig."? Sounds tautological. "Create a signature object from the R, S and V values."?

parity-crypto/src/publickey/ecdsa_signature.rs
```diff
@@ -0,0 +1,316 @@
...
+	pub fn from_rsv(r: &H256, s: &H256, v: u8) -> Self {
+		let mut sig = [0u8; 65];
+		sig[0..32].copy_from_slice(r.as_ref());
+		sig[32..64].copy_from_slice(s.as_ref());
+		sig[64] = v;
+		Signature(sig)
+	}
+
+	/// Check if this is a "low" signature.
```
dvdplm:
```suggestion
	/// Check if this is a "low" signature, i.e. a signature where `s` is lower than `0x7FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF5D576E7357A4501DDFE92F46681B20A0`
```

parity-crypto/src/publickey/ecdsa_signature.rs
```diff
@@ -0,0 +1,316 @@
...
+	pub fn is_low_s(&self) -> bool {
+		// "7FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF5D576E7357A4501DDFE92F46681B20A0"
+		const MASK: H256 = H256([
```
dvdplm: Maybe rename `MASK` to something like `LOW_SIG_THRESHOLD`?

parity-crypto/src/publickey/ecdsa_signature.rs
dvdplm: …and I for one would appreciate a remark on what we do with such "low" signatures.

parity-crypto/src/publickey/ecdsa_signature.rs
```diff
@@ -0,0 +1,316 @@
...
+			0x7F, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF,
+			0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF,
+			0x5D, 0x57, 0x6E, 0x73, 0x57, 0xA4, 0x50, 0x1D,
+			0xDF, 0xE9, 0x2F, 0x46, 0x68, 0x1B, 0x20, 0xA0,
+		]);
+		H256::from_slice(self.s()) <= MASK
+	}
+
+	/// Check if each component of the signature is in range.
```
dvdplm: State the range pls.

parity-crypto/src/publickey/ecdsa_signature.rs
```diff
@@ -0,0 +1,316 @@
...
+	pub fn is_valid(&self) -> bool {
```
dvdplm: Consider a more explicit name, e.g. `is_within_bounds` or `is_in_range`?

parity-crypto/src/publickey/ecdsa_signature.rs
```diff
@@ -0,0 +1,316 @@
...
+		// "fffffffffffffffffffffffffffffffebaaedce6af48a03bbfd25e8cd0364141"
+		const MASK: H256 = H256([
```
dvdplm: Is it really a mask? It's used as an upper bound here.

parity-crypto/src/publickey/ecdsa_signature.rs
```diff
@@ -0,0 +1,316 @@
...
+		]);
+		let r = H256::from_slice(self.r());
+		let s = H256::from_slice(self.s());
+		self.v() <= 1 &&
+			r < MASK && r >= ONE &&
+			s < MASK && s >= ONE
+	}
+}
+
+// manual implementation large arrays don't have trait impls by default.
```
dvdplm: Pls make into a `// TODO[grbIzl]`, ideally with a link to the tracking ticket.

parity-crypto/src/publickey/ecdsa_signature.rs
```diff
@@ -0,0 +1,316 @@
...
+	fn fmt(&self, f: &mut fmt::Formatter) -> Result<(), fmt::Error> {
+		write!(f, "{}", self.to_hex::<String>())
+	}
+}
+
+impl FromStr for Signature {
+	type Err = Error;
+
+	fn from_str(s: &str) -> Result<Self, Self::Err> {
+		match s.from_hex::<Vec<u8>>() {
```
dvdplm: Is this the only place where we use `rustc-hex`?

parity-crypto/src/publickey/ecdsa_signature.rs
```diff
@@ -0,0 +1,316 @@
...
+/// Signature encoded as RSV components
+#[repr(C)]
+pub struct Signature([u8; 65]);
+
+impl Signature {
+	/// Get a slice into the 'r' portion of the data.
+	pub fn r(&self) -> &[u8] {
+		&self.0[0..32]
+	}
+
```
dvdplm: Whitespace looks funky in this file (or is it just me), can you check it please?

parity-crypto/src/publickey/ecdsa_signature.rs
```diff
@@ -0,0 +1,316 @@
...
+	}
+}
+
+impl DerefMut for Signature {
+	fn deref_mut(&mut self) -> &mut Self::Target {
+		&mut self.0
+	}
+}
+
+pub fn sign(secret: &Secret, message: &Message) -> Result<Signature, Error> {
```
dvdplm: Missing docs

parity-crypto/src/publickey/ecdsa_signature.rs
```diff
@@ -0,0 +1,316 @@
...
+	let (rec_id, data) = s.serialize_compact(context);
+	let mut data_arr = [0; 65];
+
+	// no need to check if s is low, it always is
+	data_arr[0..64].copy_from_slice(&data[0..64]);
+	data_arr[64] = rec_id.to_i32() as u8;
+	Ok(Signature(data_arr))
+}
+
+pub fn verify_public(public: &Public, signature: &Signature, message: &Message) -> Result<bool, Error> {
```
dvdplm: Missing docs

parity-crypto/src/publickey/ecdsa_signature.rs
```diff
@@ -0,0 +1,316 @@
...
+
+	let publ = PublicKey::from_slice(context, &pdata)?;
+	match context.verify(&SecpMessage::from_slice(&message[..])?, &sig, &publ) {
+		Ok(_) => Ok(true),
+		Err(SecpError::IncorrectSignature) => Ok(false),
+		Err(x) => Err(Error::from(x))
+	}
+}
+
+pub fn verify_address(address: &Address, signature: &Signature, message: &Message) -> Result<bool, Error> {
```
dvdplm: Missing docs

parity-crypto/src/publickey/ecdsa_signature.rs
```diff
@@ -0,0 +1,316 @@
...
+	let public = recover(signature, message)?;
+	let recovered_address = public_to_address(&public);
+	Ok(address == &recovered_address)
+}
+
+pub fn recover(signature: &Signature, message: &Message) -> Result<Public, Error> {
```
dvdplm: Missing docs

dvdplm:
I think it's unfortunate that git history has not been preserved for these files as they migrated from the `parity-ethereum` repo. If it's too complex to do then we should at least make a not of where to look for it back in `ethkey`.

parity-crypto/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "parity-crypto"
-version = "0.4.1"
+version = "0.5.0"
```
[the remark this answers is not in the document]
ordian: We're not changing any existing API in parity-crypto, only adding new functionality, so it seems to me like a non-breaking change.

parity-crypto/src/publickey/ecdsa_signature.rs
ordian: Sorry, I might have named those while migrating to new `ethereum-types`. This is the order of the secp256k1 curve
https://github.com/bitcoin-core/secp256k1/blob/96cd94e385f64c1936abf0d1e303d12d0f5da980/src/ecdsa_impl.h#L29. Calling it an upper bounds works too.

parity-crypto/src/publickey/ec_math_utils.rs
```diff
@@ -0,0 +1,131 @@
...
+		let mut temp = [4u8; 65];
+		(&mut temp[1..65]).copy_from_slice(&public[0..64]);
+		temp
+	};
+
+	Ok(key::PublicKey::from_slice(&SECP256K1, &public_data)?)
+}
+
+fn set_public(public: &mut Public, key_public: &key::PublicKey) {
+	let key_public_serialized = key_public.serialize_vec(&SECP256K1, false);
```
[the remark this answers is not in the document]
ordian: The upstream API doesn't allow this either https://github.com/rust-bitcoin/rust-secp256k1/blob/f5c8a005f9d8948a16a0cb1c30c991df3e1f4a29/src/key.rs#L276. Unless we want to maintain our own fork ;)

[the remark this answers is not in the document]
dvdplm: So there's actually another copy in `serialize_vec()` then. Well well.

parity-crypto/Cargo.toml
ordian: Let's do this in a separate PR (cc @niklasad1), as it's non trivial.

parity-crypto/src/publickey/ec_math_utils.rs
```diff
@@ -0,0 +1,131 @@
...
+
+/// Replace public key with its negation (EC point = - EC point)
+pub fn public_negate(public: &mut Public) -> Result<(), Error> {
+	let mut key_public = to_secp256k1_public(public)?;
+	key_public.mul_assign(&SECP256K1, &key::MINUS_ONE_KEY)?;
+	set_public(public, &key_public);
+	Ok(())
+}
+
+/// Return base point of secp256k1
```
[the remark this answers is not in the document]
grbIzl: It's synonyms https://en.wikipedia.org/wiki/Elliptic-curve_cryptography#Domain_parameters

parity-crypto/src/publickey/ecdsa_signature.rs
grbIzl: This history is pretty vague already. Such format was mentioned in the yellow paper: https://ethereum.stackexchange.com/questions/55241/what-is-electrum-style-signature?rq=1 but why it was used and for which cases... it's already hidden in the mist (at least for me)

parity-crypto/src/publickey/ec_math_utils.rs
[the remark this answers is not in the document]
dvdplm: Cool. I still think we should be consistent and use the same term both for the name and the docs (where it'd be good to mention the synonym to help people that greps for the other term).

parity-crypto/src/publickey/ecdsa_signature.rs
grbIzl: This and several lines above (in fmt traits implementations)

parity-crypto/src/publickey/ecdsa_signature.rs
grbIzl: Double-checked and didn't find anything suspicious in the latest version of this file (there were spaces in the fist one)

parity-crypto/src/publickey/ec_math_utils.rs
```diff
@@ -0,0 +1,131 @@
...
+	let public_key = key::PublicKey::from_slice(&SECP256K1, &public_sec_raw)
+		.expect("constructing using predefined constants; qed");
+	let mut public = Public::default();
+	set_public(&mut public, &public_key);
+	public
+}
+
+/// Return secp256k1 elliptic curve order
+pub fn curve_order() -> U256 {
+	H256::from_slice(&CURVE_ORDER).into_uint()
```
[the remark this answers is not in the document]
grbIzl: You mean the function itself? const fn? AFAIK only const functions can be called from inside the others, but from_slice is not.

parity-crypto/src/publickey/ec_math_utils.rs
```diff
@@ -0,0 +1,131 @@
...
+}
+
+fn to_secp256k1_public(public: &Public) -> Result<key::PublicKey, Error> {
+	let public_data = {
+		let mut temp = [4u8; 65];
```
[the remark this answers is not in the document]
grbIzl: Sorry, but I don't get it. Why to set all bits to zero here?

parity-crypto/src/publickey/ec_math_utils.rs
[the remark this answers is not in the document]
dvdplm: No, what was getting at was: this code is constant; sure it can't be just `const` but maybe `lazy_static!` can be used to reduce the calls here?

parity-crypto/src/publickey/ec_math_utils.rs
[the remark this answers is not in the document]
dvdplm: Apologies for being unclear, I was thinking out loud. Maybe, I was thinking, it's faster on the assembly level to initialize a chunk of memory to zero than to another value. Here you initialize the array with `4` because that's what you need in `t[0]` and then move on to copy over the rest of the array. If I'm wrong and initializing the array with `4` is just as fast as with `0` then obviously your code here is good. I was just curious if you knew if there is such a fast-path in hardware for zero-initialized memory, and if yes, if it actually makes any difference.

parity-crypto/src/publickey/ecdsa_signature.rs
dvdplm: :/

Seems like it's used in `rpc` and `private-tx` (and tests).

parity-crypto/src/publickey/ecdsa_signature.rs
dvdplm: Let's use their comment, "Group order for secp256k1 defined as 'n' in "Standards for Efficient Cryptography" (SEC2) 2.7.1; used here as the upper bound for a valid (r, s, v) tuple" (or something like that)?

parity-crypto/src/publickey/ec_math_utils.rs
grbIzl: I've added one more test for negation operation, that makes sense for me. For other methods I could not come up with anything better, than check returning value against its expected hardcoded const. So I've left them for now

parity-crypto/src/publickey/ecdh.rs
grbIzl: I've renamed this method in order to reflect its actual work.

parity-crypto/src/publickey/ecdh.rs
grbIzl: Benchmark is added.

parity-crypto/src/publickey/ecdsa_signature.rs
grbIzl: I don't agree with your suggestion, this method is aimed to check signature's validity at the first place. Checking ranges is just its implementation

parity-crypto/src/publickey/ec_math_utils.rs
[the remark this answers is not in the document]
grbIzl: I'm aware of only std::mem::zeroed, but a) not sure, how fast(er) it is; b) it's unsafe and not recommended for usage.

parity-crypto/src/publickey/ec_math_utils.rs
```diff
@@ -0,0 +1,131 @@
...
+	set_public(public, &key_public);
+	Ok(())
+}
+
+/// Return base point of secp256k1
+pub fn generation_point() -> Public {
+	let mut public_sec_raw = [0u8; 65];
+	public_sec_raw[0] = 4;
+	public_sec_raw[1..33].copy_from_slice(&GENERATOR_X);
+	public_sec_raw[33..65].copy_from_slice(&GENERATOR_Y);
```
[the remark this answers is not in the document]
grbIzl: Sorry for my illiteracy, but I don't know, how to assemble array from the other arrays at the compile time.

grbIzl:
> I think it's unfortunate that git history has not been preserved for these files as they migrated from the `parity-ethereum` repo. If it's too complex to do then we should at least make a note of where to look for it back in `ethkey`.

It's a common problem for such case, when files are moved to the other repo. So I've put a comment, but not sure, how helpful it will be

parity-crypto/src/publickey/ec_math_utils.rs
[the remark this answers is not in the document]
dvdplm: I was thinking something like this:

```rust
const GENERATOR_X: [u8; 32] = [1,3,2,4,9, … …];
const GENERATOR_Y: [u8; 32] = [2,4,3,8,8, … …];
const BASE_POINT_BYTES: [u8; 65] = [4, 1,3,2,4,9, … …, 2,4,3,8,8, … …];

pub fn generation_point() -> Public {
  let mut public_sec_raw = [0u8; 65];
  public_sec_raw[0..65].copy_from_slice(&BASE_POINT_BYTES);
  …
}
```

…but I'm probably missing something! :)

parity-crypto/src/publickey/ec_math_utils.rs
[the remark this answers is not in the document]
dvdplm: If what I suggest works it would be a compiler optimization.

parity-crypto/src/publickey/ec_math_utils.rs
dvdplm: > For other methods I could not come up with anything better, than check returning value against its expected hardcoded const

Isn't that ok though? Like, these are unit tests, they don't have to be "interesting" in any way, just make sure that the calls do what is expected? Or what am I missing?

parity-crypto/src/publickey/ec_math_utils.rs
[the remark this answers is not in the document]
grbIzl: A, you mean this. Got it.

parity-crypto/src/publickey/ec_math_utils.rs
grbIzl: Ok. Added such tests for other methods

parity-crypto/Cargo.toml
```diff
@@ -24,9 +27,15 @@ aes = "0.3.2"
...
 subtle = "2.1"
 zeroize = "0.9.1"
 
 [dev-dependencies]
 criterion = "0.2"
 hex-literal = "0.2"
+
+[features]
+default = []
+publickey = ["eth-secp256k1", "lazy_static", "ethereum-types"]
```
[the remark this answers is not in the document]
ordian: Does it make sense to name this feature `secp256k1`? `publickey` seems too general

parity-crypto/src/publickey/ec_math_utils.rs
```diff
@@ -18,10 +18,25 @@
...
 use super::{SECP256K1, Public, Secret, Error};
 use secp256k1::key;
-use secp256k1::constants::{GENERATOR_X, GENERATOR_Y, CURVE_ORDER as SECP256K1_CURVE_ORDER};
+use secp256k1::constants::{CURVE_ORDER as SECP256K1_CURVE_ORDER};
 use ethereum_types::{BigEndianHash as _, U256, H256};
 use lazy_static::lazy_static;
 
+/// Generation point array combined from X and Y coordinates
+pub const BASE_POINT_BYTES: [u8; 65] = [
+	0x4,
```
dvdplm: a comment documenting this `0x4` too would be good.

parity-crypto/src/publickey/ec_math_utils.rs
```diff
@@ -148,4 +149,24 @@ mod tests {
...
+	#[test]
+	fn known_public_is_valid() {
+		let public = Random.generate().unwrap().public().clone();
+		assert!(public_is_valid(&public));
+	}
+
+	#[test]
+	fn generation_point_expected() {
+		let point = generation_point();
+		assert_eq!(format!("{:x}", point), "79be667ef9dcbbac55a06295ce870b07029bfcdb2dce28d959f2815b16f81798483ada7726a3c4655da4fbfc0e1108a8fd17b448a68554199c47d08ffb10d4b8");
```
dvdplm: please add a comment on where the generation point comes from?

parity-crypto/src/publickey/ec_math_utils.rs
```diff
@@ -148,4 +149,24 @@ mod tests {
...
+	}
+
+	#[test]
+	fn public_multiplication_verification() {
+		let secret = Secret::from_str("a100df7a048e50ed308ea696dc600215098141cb391e9527329df289f9383f65").unwrap();
+		let mut public = generation_point();
+		public_mul_secret(&mut public, &secret).unwrap();
+		assert_eq!(format!("{:x}", public), "8ce0db0b0359ffc5866ba61903cc2518c3675ef2cf380a7e54bde7ea20e6fa1ab45b7617346cd11b7610001ee6ae5b0155c41cad9527cbcdff44ec67848943a4");
```
dvdplm: If this value is from a spec somewhere it'd be great to document that. :)

parity-crypto/Cargo.toml
```diff
@@ -11,7 +11,7 @@ edition = "2018"
 [[bench]]
 name = "bench"
 harness = false
-
+required-features = ["publickey"]
```
dvdplm: 👍

dvdplm:
> It's a common problem for such case, when files are moved to the other repo.

It is. The trick is to using something like `git subtree extract`. I should have spoken up when you started this work. :/

parity-crypto/src/publickey/secret_key.rs
```diff
@@ -0,0 +1,304 @@
...
+		let secret = key::SecretKey::from_slice(&super::SECP256K1, key)?;
+		Ok(secret.into())
+	}
+
+	/// Checks validity of this key.
+	pub fn check_validity(&self) -> Result<(), Error> {
+		self.to_secp256k1_secret().map(|_| ())
+	}
+
+	/// Wrapper over to hex conversion
```
dvdplm:
```suggestion
	/// Wrapper over hex conversion
```

parity-crypto/src/publickey/mod.rs
```diff
@@ -0,0 +1,59 @@
...
+pub use ethereum_types::{Address, Public};
+pub type Message = H256;
+
+lazy_static! {
+	pub static ref SECP256K1: secp256k1::Secp256k1 = secp256k1::Secp256k1::new();
+}
+
+/// Uninstantiatable error type for infallible generators.
+#[derive(Debug)]
+pub enum Void {}
```
dvdplm: Now that `Infallible` is stable, can we switch to that?

parity-crypto/src/publickey/mod.rs
```diff
@@ -0,0 +1,59 @@
...
+// Parity Ethereum is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Parity Ethereum.  If not, see <http://www.gnu.org/licenses/>.
+
+//! Submodule of crypto utils for working with public key crypto primitives
+//! All functionality moved from ethkey module in parity ethereum repository
```
dvdplm:
```suggestion
//! If you are looking for git history please refer to the `ethkey` crate in the `parity-ethereum` repository.
```

parity-crypto/src/publickey/keypair_generator.rs
```diff
@@ -0,0 +1,46 @@
...
+
+// Parity Ethereum is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Parity Ethereum.  If not, see <http://www.gnu.org/licenses/>.
+
+//! Key pair generator's description
```
dvdplm:
```suggestion
//! Random key pair generator. Relies on the secp256k1 C-library to generate random data.
```

parity-crypto/src/publickey/keypair.rs
```diff
@@ -0,0 +1,116 @@
...
+
+	pub fn from_keypair(sec: key::SecretKey, publ: key::PublicKey) -> Self {
+		let context = &SECP256K1;
+		let serialized = publ.serialize_vec(context, false);
+		let secret = Secret::from(sec);
+		let mut public = Public::default();
+		public.as_bytes_mut().copy_from_slice(&serialized[1..65]);
+
+		KeyPair {
+			secret: secret,
```
dvdplm:
```suggestion
			secret,
```

parity-crypto/src/publickey/keypair.rs
```diff
@@ -0,0 +1,116 @@
...
+			public: public,
```
dvdplm:
```suggestion
			public,
```

parity-crypto/src/publickey/keypair.rs
```diff
@@ -0,0 +1,116 @@
...
+
+		let keypair = KeyPair {
+			secret: secret,
+			public: public,
+		};
+
+		Ok(keypair)
+	}
+
+	pub fn from_secret_slice(slice: &[u8]) -> Result<KeyPair, Error> {
```
dvdplm: Missing docs

parity-crypto/src/publickey/keypair.rs
```diff
@@ -0,0 +1,116 @@
...
+		Self::from_secret(Secret::import_key(slice)?)
+	}
+
+	pub fn from_keypair(sec: key::SecretKey, publ: key::PublicKey) -> Self {
```
dvdplm: Missing docs.

parity-crypto/src/publickey/keypair.rs
```diff
@@ -0,0 +1,116 @@
...
+
+// Parity Ethereum is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Parity Ethereum.  If not, see <http://www.gnu.org/licenses/>.
+
+//! Key pair (public + secrect) description
```
dvdplm: Missing docs

parity-crypto/src/publickey/extended_keys.rs
```diff
@@ -0,0 +1,521 @@
...
+
+// Parity Ethereum is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Parity Ethereum.  If not, see <http://www.gnu.org/licenses/>.
+
+//! Secret, public keys extended with the entropy (aka chain code), that allows further keys derivation
```
dvdplm:
```suggestion
//! Secret, public keys extended with the entropy (aka chain code), that allows further key derivation.
```

parity-crypto/src/publickey/error.rs
```diff
@@ -0,0 +1,98 @@
...
+// You should have received a copy of the GNU General Public License
+// along with Parity Ethereum.  If not, see <http://www.gnu.org/licenses/>.
+
+//! Module specific errors
+
+use std::{fmt, result, error::Error as StdError};
+use crate::error::SymmError;
+
+#[derive(Debug)]
+pub enum Error {
```
dvdplm: Missing docs

parity-crypto/src/publickey/ecies.rs
```diff
@@ -0,0 +1,138 @@
...
+	Ok(msg)
+}
+
+/// Decrypt a message with a secret key, checking HMAC for ciphertext
+/// and authenticated data validity.
+pub fn decrypt(secret: &Secret, auth_data: &[u8], encrypted: &[u8]) -> Result<Vec<u8>, Error> {
+	const META_LEN: usize = 1 + 64 + 16 + 32;
+	let enc_version = encrypted[0];
+	if encrypted.len() < META_LEN || enc_version < 2 || enc_version > 4 {
+		return Err(Error::InvalidMessage); //invalid message: publickey
```
dvdplm: Not sure I understand the comment here. Can you clarify?

parity-crypto/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "parity-crypto"
-version = "0.4.1"
+version = "0.4.2"
```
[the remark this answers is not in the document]
ordian: I see where you're coming from and don't have a strong opinion on this. Having a minor release makes it easier to ensure other crates could use the latest version of parity-crypto (even via transitive dependencies), but that's a minor concern.

parity-crypto/src/publickey/ec_math_utils.rs
grbIzl: It's just a randomly created secret

parity-crypto/src/publickey/secret_key.rs
```diff
@@ -0,0 +1,304 @@
...
+	pub fn inv(&mut self) -> Result<(), Error> {
+		let mut key_secret = self.to_secp256k1_secret()?;
+		key_secret.inv_assign(&SECP256K1)?;
+
+		*self = key_secret.into();
+		Ok(())
+	}
+
+	/// Compute power of secret key inplace (secret ^ pow).
+	/// This function is not intended to be used with large powers.
```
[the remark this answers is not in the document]
grbIzl: I've removed this comment as it's misleading. The only limitation here is pow (usize). The multiplication on elliptic curves doesn't have any limitations

parity-crypto/src/publickey/ecies.rs
grbIzl: Comment removed

parity-crypto/Cargo.toml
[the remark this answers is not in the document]
grbIzl: I agree in general, but this exact name (secp256k1) seems not the best option, because we already have eth-secp256k1 feature and it will be two similar names. Naming is the hardest part of coding :-(

grbIzl:
> So what are the next steps after this? A PR to `parity-ethereum` to use this I guess?

Yep.

dvdplm:
k, I'm putting this `onice` so it doesn't get merged before it's time.

parity-crypto/Cargo.toml
```diff
@@ -11,10 +11,13 @@ edition = "2018"
 [[bench]]
 name = "bench"
 harness = false
-
+required-features = ["publickey"]
 
 [dependencies]
 tiny-keccak = "1.4"
+eth-secp256k1 = { git = "https://github.com/paritytech/rust-secp256k1", rev = "a96ad75", optional = true }
```
ordian:
```suggestion
eth-secp256k1 = { git = "https://github.com/paritytech/rust-secp256k1", rev = "72c93ab = true }
```
another update

niklasad1:
@grbIzl @dvdplm let's publish a new version of `parity-crypto` so it can be used by https://github.com/paritytech/parity-ethereum/pull/11174?

grbIzl:
> @grbIzl @dvdplm let's publish a new version of `parity-crypto` so it can be used by [paritytech/parity-ethereum#11174](https://github.com/paritytech/parity-ethereum/pull/11174)?

Working on it. Right now bumped into https://github.com/paritytech/rust-secp256k1/pull/25

PR #212, opened 2019-09-03
[parity-bytes] alloc crate has stabilized

koushiro:


dvdplm:
> "Minimum supported rust version" policy

Latest stable is fine imo.

PR #218, opened 2019-09-09
Find PendingIterator in Transaction Pool

ngotchac:
When encountering a `Stale` transaction in the `PendingIterator`, we should try to use the next best transaction for this sender.

### DIFF
--- a/transaction-pool/src/pool.rs
+++ b/transaction-pool/src/pool.rs
@@ -615,7 +615,7 @@ impl<'a, T, R, S, L> Iterator for PendingIterator<'a, T, R, S, L> where
 			// Add the next best sender's transaction when applicable
 			match tx_state {
 				Readiness::Ready | Readiness::Stale => {
-					// retrieve next one from that sender.
+					// retrieve next one from the same sender.
 					let next = self.pool.transactions
 						.get(best.transaction.sender())
 						.and_then(|s| s.find_next(&best.transaction, &self.pool.scoring));
@@ -625,12 +625,12 @@ impl<'a, T, R, S, L> Iterator for PendingIterator<'a, T, R, S, L> where
 				},
 				_ => (),
 			}
-			match tx_state {
-				Readiness::Ready => {
-					return Some(best.transaction.transaction)
-				},
-				_ => trace!("[{:?}] Ignoring {:?} transaction.", best.transaction.hash(), tx_state),
+
+			if tx_state == Readiness::Ready {
+				return Some(best.transaction.transaction)
 			}
+
+			trace!("[{:?}] Ignoring {:?} transaction.", best.transaction.hash(), tx_state);
 		}
 
 		None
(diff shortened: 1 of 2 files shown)


transaction-pool/src/pool.rs
```diff
@@ -611,19 +611,25 @@ impl<'a, T, R, S, L> Iterator for PendingIterator<'a, T, R, S, L> where
...
+			// Add the next best sender's transaction when applicable
+			match tx_state {
+				Readiness::Ready | Readiness::Stale => {
 					// retrieve next one from that sender.
 					let next = self.pool.transactions
 						.get(best.transaction.sender())
 						.and_then(|s| s.find_next(&best.transaction, &self.pool.scoring));
 					if let Some((score, tx)) = next {
 						self.best_transactions.insert(ScoreWithRef::new(score, tx));
 					}
```
tomusdrw: Can't we do:
```
if tx_state == Readiness::Ready {
   return Some(best.transaction.transaction);
}
```
instead of this additional match?

transaction-pool/src/pool.rs
```diff
@@ -611,19 +611,25 @@ impl<'a, T, R, S, L> Iterator for PendingIterator<'a, T, R, S, L> where
 				self.best_transactions.take(&best).expect("Just taken from iterator; qed")
 			};
 
-			match self.ready.is_ready(&best.transaction) {
-				Readiness::Ready => {
+			let tx_state = self.ready.is_ready(&best.transaction);
+			// Add the next best sender's transaction when applicable
+			match tx_state {
+				Readiness::Ready | Readiness::Stale => {
 					// retrieve next one from that sender.
```
dvdplm:
```suggestion
					// retrieve next one from the same sender.
```

transaction-pool/src/pool.rs
```diff
@@ -625,12 +625,12 @@ impl<'a, T, R, S, L> Iterator for PendingIterator<'a, T, R, S, L> where
 				},
 				_ => (),
 			}
-			match tx_state {
-				Readiness::Ready => {
-					return Some(best.transaction.transaction)
-				},
-				_ => trace!("[{:?}] Ignoring {:?} transaction.", best.transaction.hash(), tx_state),
+
+			if tx_state == Readiness::Ready {
```
tomusdrw: I've actually meant to put the if inside previous match and leave the trace in `_` pattern, but this is fine as well.

Just got an idea that for readability we could also change both this `if` and `match` above to `if let`, but it's up to you.

ngotchac: I thought it would be clearer to have one match when we need to find the next transactions, and another one when we need to return the transaction, but I'm open to anything :)

tomusdrw: My reasoning was that it's easier to see everything that happens in particular case(ses) in a single block, cause now you fall-through and need to analyze multiple blocks to think about what happens. But as I said, I'm not strong on this either.

PR #222, opened 2019-09-22
[kvdb-memorydb] Migrated code to 2018 edition, updated parking_lot

expenses:
See #143.

### DIFF
--- a/kvdb-memorydb/Cargo.toml
+++ b/kvdb-memorydb/Cargo.toml
@@ -1,11 +1,12 @@
 [package]
 name = "kvdb-memorydb"
-version = "0.1.0"
+version = "0.1.1"
 authors = ["Parity Technologies <admin@parity.io>"]
 repository = "https://github.com/paritytech/parity-common"
 description = "A key-value in-memory database that implements the  `KeyValueDB` trait"
 license = "GPL-3.0"
+edition = "2018"
 
 [dependencies]
-parking_lot = "0.6"
+parking_lot = "0.9"
 kvdb = { version = "0.1", path = "../kvdb" }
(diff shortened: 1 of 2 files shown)


kvdb-memorydb/Cargo.toml
```diff
@@ -1,11 +1,12 @@
...
+version = "0.1.1"
 authors = ["Parity Technologies <admin@parity.io>"]
 repository = "https://github.com/paritytech/parity-common"
 description = "A key-value in-memory database that implements the  `KeyValueDB` trait"
 license = "GPL-3.0"
+edition = "2018"
 
 [dependencies]
-parking_lot = "0.6"
+parking_lot = "0.9"
```
dvdplm: We're on `0.8` in `parity-ethereum` and I have a hunch that this could break things. If you feel like trying this change against the master branch over there, that'd be fantastic.

expenses:
There are some breakages such as
```
error[E0277]: the trait bound `kvdb_memorydb::InMemory: kvdb::KeyValueDB` is not satisfied
   --> miner/local-store/src/lib.rs:219:30
    |
219 |             let store = super::create(db.clone(), None, Dummy(vec![]));
    |                                       ^^^^^^^^^^ the trait `kvdb::KeyValueDB` is not implemented for `kvdb_memorydb::InMemory`
    |
    = note: required for the cast to the object type `dyn kvdb::KeyValueDB`
```

Perhaps this should be version `0.2.0`?

dvdplm:
@expenses you mean change this crate to v0.2.0? If we upgrade parking_lot to a breaking version then yes. 
I haven’t looked closely at the error you posted yet but it could be we need to update all of parity-ethereum at once, or is that what you did perhaps?

The best way imo would be to revert the parking_lot upgrade and get this merged, then we deal with upgrading the reps separately.

And thanks again for your work here! :)

expenses:
Sure, I'll do that.

PR #230, opened 2019-09-30
Use default Xenial dist

hugovk:
Xenial is now the default:

* https://blog.travis-ci.com/2019-04-15-xenial-default-build-environment


hugovk:
I guess they hadn't finished the rollout yet.

PR #231, opened 2019-10-01
[parity-util-mem] migrate code to 2018 edition

iamsauravsharma:
Migrate parity-util-mem to 2018 edition
Part of #143


iamsauravsharma:
[clabot:check]

parity-util-mem/Cargo.toml
```diff
@@ -5,6 +5,7 @@ authors = ["Parity Technologies <admin@parity.io>"]
 repository = "https://github.com/paritytech/parity-common"
 description = "Collection of memory related utilities"
 license = "GPL-3.0"
+edition = "2018"
```
niklasad1: don't we need `[lib] proc-macro = true` in order make the procedural macro `MallocSizeOf` to work? 

Otherwise, I think we need to `#[macro_use] extern crate parity_util_mem as malloc_size_of;` which is annoying IMHO.

cheme:
@iamsauravsharma , thanks for going through it.
Things seems fine to me, except the ci was lacking a use case, I updated things in this commit : https://github.com/cheme/parity-common/commit/d65d43eb63e64afa1935101bcd780ecd9e7d6b22 , see if you can add it to your PR (I think it should be removed, but still updated it).

dvdplm:
> (I think it should be removed, but still updated it).

Not sure I understand what you mean here: what is it that should be removed?

cheme:
get_malloc_size_src.sh and slim_malloc_size_of.patch (the comment with github commit to servo commit should be enough). It was probably me that wanted to keep it in the first place, but time passing I realize it does not really have any value.

parity-util-mem/Cargo.toml
cheme: I just tested with 'proc-macro = true', it is only for procedural crate, here we reexport procedural crate, so it was a no success for me :( . I will therefore approve the pr as his.

niklasad1: Ok, I see thanks @cheme :P

PR #232, opened 2019-10-01
[trace-time] migrate code to 2018 edition

iamsauravsharma:
Part of #143

### DIFF
--- a/trace-time/src/lib.rs
+++ b/trace-time/src/lib.rs
@@ -16,16 +16,14 @@
 
 //! Performance timer with logging
 
-#[macro_use]
-extern crate log;
-
+use log::trace;
 use std::time::Instant;
 
 #[macro_export]
 macro_rules! trace_time {
 	($name: expr) => {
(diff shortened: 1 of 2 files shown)


trace-time/src/lib.rs
```diff
@@ -48,8 +46,7 @@ impl PerfTimer {
 impl Drop for PerfTimer {
 	fn drop(&mut self) {
 		let elapsed = self.start.elapsed();
-		let ms = elapsed.subsec_nanos() as f32 / 1_000_000.0 +
-				 elapsed.as_secs() as f32 * 1_000.0;
+		let ms = elapsed.subsec_nanos() as f32 / 1_000_000.0 + elapsed.as_secs() as f32 * 1_000.0;
```
niklasad1: use https://doc.rust-lang.org/std/time/struct.Duration.html#method.as_millis instead, introduced in rust 1.33

```suggestion
		let ms = elapsed.as_millis();
```

PR #236, opened 2019-10-08
Upgrade depenencies

DemiMarie-parity:


dvdplm:
A lot of funny looking lifetime annotations appeared here, why is that? And while explicit versions down to the patch level is fine I am not sure they are strictly needed, can you elaborate a bit on what motivates these changes?

niklasad1:
@dvdplm I think it fixes a lint/warning part of the `rust_2018_idioms` which is `warning: hidden lifetime parameters in types are deprecated`

fixed-hash/src/hash.rs
```diff
@@ -526,13 +526,6 @@ macro_rules! impl_rand_for_fixed_hash {
 				ret.randomize_using(rng);
 				ret
 			}
-
-			/// Create a new hash with cryptographically random content.
-			pub fn random() -> Self {
```
niklasad1: why remove? It is a breaking change

parity-crypto/benches/bench.rs
```diff
@@ -15,7 +15,7 @@
 // along with Parity.  If not, see <http://www.gnu.org/licenses/>.
 
 
-extern crate parity_crypto;
+use parity_crypto;
```
bkchr: This is not required.

kvdb-rocksdb/src/lib.rs
```diff
@@ -703,7 +703,7 @@ impl Drop for Database {
 
 #[cfg(test)]
 mod tests {
-	extern crate tempdir;
+	use tempdir;
```
bkchr: Not required.

Just remove the `use self::tempdir` below and just use `tempdir::`

parity-crypto/benches/bench.rs
DemiMarie-parity: Agreed.  This was done automatically.

fixed-hash/src/hash.rs
DemiMarie-parity: `rustc` warned about it being unused.

dvdplm: How can cargo fix know that a `pub` item is not used?

PR #240, opened 2019-10-13
Fix old syntax in the README of crate uint

eisterman:
The syntax actually reported is not working and is different from that present in the test and the examples folders.

### DIFF
--- a/uint/README.md
+++ b/uint/README.md
@@ -16,7 +16,13 @@ In your `Cargo.toml` paste
 uint = "0.8"
 ```
 
-Put in the header of your main file
+Import the macro
+
+```
+use uint::construct_uint;
+```
+
+If you're using pre-edition Rust in your main file
 
 ```
 #[macro_use]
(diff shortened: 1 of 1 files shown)


eisterman:
[clabot:check]

uint/README.md
```diff
@@ -16,11 +16,20 @@ In your `Cargo.toml` paste
 uint = "0.8"
 ```
 
+Put in the header of your main file
```
dvdplm: I'd prefer we show the Edition 2018 syntax first, and I'm not sure about the "header" part, it's an inflated word.

So maybe:

```
Import the macro:

use uint::construct_uint;

If you're using pre-edition Rust in your project:

#[macro_use]
extern crate uint;
```

niklasad1: I agree but that is really messy currently because the inner macros are not `rust2018 idiomatic`  i.e, all must be exported which makes you need to do

```rust
use uint::{very many imports / * }
```

eisterman: I think the same thing, with that pull-request the syntax `use uint::construct_uint` will work correctly

PR #243, opened 2019-10-21
Fix off-by-one.

tomusdrw:


tomusdrw:
@ordian Indeed Parity Ethereum is following the jsonrpc spec from here: https://github.com/ethereum/wiki/wiki/JSON-RPC#hex-value-encoding and it clearly states what values are valid and what not.
`impl-serde` though is a bit more relaxed in this.

primitive-types/impls/serde/src/serialize.rs
```diff
@@ -105,7 +105,7 @@ pub fn deserialize<'de, D>(deserializer: D) -> Result<Vec<u8>, D::Error> where
 
 			let bytes_len = v.len() - 2;
 			let mut modulus = bytes_len % 2;
-			let mut bytes = vec![0u8; bytes_len / 2 + 1];
+			let mut bytes = vec![0u8; (bytes_len + 1) / 2];
```
niklasad1: why bother with this complex logic? 
Isn't better to use `Vec::new() / Vec::with_capacity()` and Vec::push?

tomusdrw: Well, it's a performance optimization (hopefuly measureable), which is justified, since this code might be on a hot path.
Also this logic is not that complex imho, seems it was just untested properly.

dvdplm: I for one kind of agree with Niklas here: it's a bit odd and I'm surprised to learn it's faster to init a `vec ` with `vec![0u8; 123]` than `Vec::with_capacity(123)`.

(No action needed here, I'm just curious)

niklasad1: I'm about to add benches for `Bytes` to be continued :P

tomusdrw: Wow, thanks for making me realize this is actually a `vec`. I was pretty sure that we just allocate an array here (but obviously we don't cause it's not a constant size)
I'm really curious about the performance indeed, as just using `push` instead of direct access seems way less error prone, and `with_capacity` should ensure that there are no re-allocations.

dvdplm: I really hope there's no difference in performance! :)

PR #244, opened 2019-10-22
export FromDecStrErr

WyseNynja:
My project needs to be able to convert from uint's error type to my project's error type. This should make it possible to do this without needing to add uint as a direct dependency of my project.

### DIFF
--- a/ethereum-types/src/lib.rs
+++ b/ethereum-types/src/lib.rs
@@ -5,7 +5,7 @@
 mod hash;
 mod uint;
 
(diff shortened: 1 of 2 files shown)


WyseNynja:
[clabot:check]

WyseNynja:
I'll get this actually working with my project and then I'll update this PR.

WyseNynja:
Has this been released yet? I'd like to submit a PR to https://github.com/tomusdrw/rust-web3, but I don't want them to have to use master.

PR #250, opened 2019-10-26
Release impl-rlp 0.2.1

sorpaas:
The newest commit of `impl-rlp` brings in `no_std` support, but it's not released.


dvdplm:
> The newest commit

Which commit was that if you don't mind?

niklasad1:
https://github.com/paritytech/parity-common/pull/206

dvdplm:
published `v0.2.1`

PR #290, opened 2019-12-21
Move and extend impls for locking primitives

NikVolf:
- they probably should be under "std" feature flag, not "ethereum"
- `std::sync::RwLock` impl was missing


NikVolf:
@ordian addressed

NikVolf:
@dvdplm pls add a 2nd review?

NikVolf:
@dvdplm I added comment, let me know if there is anything else I can do

dvdplm:
> let me know if there is anything else I can do

We'll need to make a proper changelog entry but as there are several PRs in-flight atm I guess it's fine to do one pre-release PR where we write this down along all the rest.

PR #300, opened 2020-01-02
Use "impl-trait-for-tuples" for tuples in parity-util-mem

NikVolf:


parity-util-mem/src/malloc_size.rs
```diff
@@ -261,42 +261,12 @@ impl<T: MallocSizeOf + ?Sized> MallocSizeOf for Box<T> {
...
-	T2: MallocSizeOf,
-	T3: MallocSizeOf,
-	T4: MallocSizeOf,
-{
+#[impl_trait_for_tuples::impl_for_tuples(12)]
+impl MallocSizeOf for Tuple {
 	fn size_of(&self, ops: &mut MallocSizeOfOps) -> usize {
-		self.0.size_of(ops) + self.1.size_of(ops) + self.2.size_of(ops) + self.3.size_of(ops)
+		let mut result = 0;
+		for_tuples!( #( result += Tuple.size_of(ops); )* );
```
bkchr: The full function can be written like:

```
for_tuples!( #( Tuple.size_of(ops) )+* )
```

NikVolf: This is first what I tried - does not work, probably because of `()` variant is also expanded, producing `()` return value

bkchr: Ahh yeah. Do we need the `()` case? If not `impl_for_tuple(1, 12)`

NikVolf: We do need it

PR #318, opened 2020-01-26
uint: make zero const fn

vorot93:
Since `zero` is special in that it is the usual initialization default, I figured I could unblock `const fn new` for structs that use uint types.


vorot93:
[clabot:check]

vorot93:
[clabot:check]

PR #319, opened 2020-01-27
Expand const fn coverage

vorot93:
Follow up to #318

### DIFF
--- a/fixed-hash/src/hash.rs
+++ b/fixed-hash/src/hash.rs
@@ -106,16 +106,19 @@ macro_rules! construct_fixed_hash {
 
 		impl $name {
 			/// Returns a new fixed hash where all bits are set to the given byte.
+			#[inline]
 			pub const fn repeat_byte(byte: u8) -> $name {
 				$name([byte; $n_bytes])
 			}
 
 			/// Returns a new zero-initialized fixed hash.
+			#[inline]
 			pub const fn zero() -> $name {
 				$name::repeat_byte(0u8)
 			}
 
 			/// Returns the size of this hash in bytes.
+			#[inline]
 			pub const fn len_bytes() -> usize {
 				$n_bytes
 			}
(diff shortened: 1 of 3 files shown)


dvdplm:
What are your thoughts on what criteria to use for deciding what should be `const` and what shouldn't? E.g. in `uint` there are a few methods left non-const: `low_u128()`, `bit()`, `split()`, `split_u128()` – what was the reasoning there?

Another question I have is about the removed `#[inline]`s: I thought `const fn` meant it __can__ be transformed to a constant at compile time, but is not guaranteed to be. In other words, that in some code it'd be constified but other times called as a normal function. Am I mistaken?

vorot93:
@dvdplm my idea is to make pretty much everything `const fn` unless it is not allowed to be such. That means that the most basic of methods with no boolean logic or mutation can become `const fn` today.

re: inline - we would need a compiler expert here, but I'd expect any function that uses `const fn` to be classified by internal heuristics as potentially inline-able. This is not something I'm completely sure about though and can restore the attribute.

dvdplm:
> I missed the methods you listed and will make them `const fn` right away.

Got it, cool. It would be great to have a tool/lint to help us find code that can be constified. :)

dvdplm:
> re: inline - we would need a compiler expert here, but I'd expect any function that uses `const fn` to be classified by internal heuristics as potentially inline-able. This is not something I'm completely sure about though and can restore the attribute.

Reading up a little about this, my understanding is that a `const fn` implies a guarantee that the function is callable in a const context; when called in a non-const context the `const fn` is called like a normal function. E.g. given:

```
const fn is_even(x: usize) -> bool { x % 2 == 0 }
```

…then `is_even(42)` is guaranteed to be const, but if I have `fn random() -> usize` somewhere and call `is_even(random())` then it is not folded.

My conclusion is that we want to keep the `#[inline]`.

vorot93:
@dvdplm thank you for the write-up. I have added back `#[inline]` in the last commit.

vorot93:
By the way, there is an actual clippy lint for potential const fn now :)
https://rust-lang.github.io/rust-clippy/master/index.html#missing_const_for_fn

PR #339, opened 2020-02-09
Disable cache if explicit memory budget=0 passed

### LINKED ISSUE #330: No-cache configuration for kvdb-rocksdb
For benchmarking, transaction-weight calculation and exploring cold-cache attacks, it is useful to create rocksdb with zero read cache.

Looks like it is supported in the rocksdb itself with this:
`table_options.no_block_cache = true;`

correct me if I'm wrong

NikVolf:
No need to create LRUCache cache and set other cache options if user want low memory profile and thus no cache!

Also fixes #330

### DIFF
--- a/kvdb-rocksdb/src/lib.rs
+++ b/kvdb-rocksdb/src/lib.rs
@@ -236,7 +236,7 @@ impl MallocSizeOf for DBAndColumns {
 	fn size_of(&self, ops: &mut parity_util_mem::MallocSizeOfOps) -> usize {
 		let mut total = self.column_names.size_of(ops)
 			// we have at least one column always, so we can call property on it
-			+ self.static_property_or_warn(0, "rocksdb.block-cache-usage");
+			+ self.maybe_property(0, "rocksdb.block-cache-usage").unwrap_or(0);
 
 		for v in 0..self.column_names.len() {
 			total += self.static_property_or_warn(v, "rocksdb.estimate-table-readers-mem");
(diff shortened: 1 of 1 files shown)


kvdb-rocksdb/src/lib.rs
```diff
@@ -236,7 +236,7 @@ impl MallocSizeOf for DBAndColumns {
 	fn size_of(&self, ops: &mut parity_util_mem::MallocSizeOfOps) -> usize {
 		let mut total = self.column_names.size_of(ops)
 			// we have at least one column always, so we can call property on it
-			+ self.static_property_or_warn(0, "rocksdb.block-cache-usage");
```
NikVolf: this actually produces warning when block cache is disabled 🤷‍♀️

kvdb-rocksdb/src/lib.rs
```diff
@@ -261,6 +261,10 @@ impl DBAndColumns {
 			}
 		}
 	}
+
+	fn maybe_property(&self, col: usize, prop: &str) -> Option<usize> {
```
dvdplm: Just a personal opinion but I think I'd prefer to keep this logic in the original place; seems like overkill to have its own method.

NikVolf: what original place?

dvdplm: Sorry, I meant: this method is called only from one place, so this one-liner doesn't buy us much in terms of cleaner code.

PR #423, opened 2020-08-24
uint/src/uint.rs: Fix doc comment typo

mxinden:


ordian (approved):
Thanks for catching this copy-paste error!

mxinden:
I doubt this is worth mentioning in the release notes or even worth bumping the version. Let me know if you think otherwise.

dvdplm (approved):
oops! :)

PR #425, opened 2020-08-26
Fix panic in FromStr implementation 

e00E:
The current implementation panics if the hex is string is too big in
`from_big_endian`.

Also removed duplicate tests that were probably a copy paste error.


ordian (approved):
Good catch, thank you!

dvdplm (approved):

PR #429, opened 2020-09-19
Support for tlp decodable with removed leading zeros

crystalin:
It adds support for RLP encoded values of expected length (like H256) trimmed of their leading zeros.


crystalin:
Closing this. The current behavior is the expected behavior for H256

PR #431, opened 2020-10-08
rlp: u128 support

vorot93:


dvdplm:
A description would be good: why do we need this?

ordian (approved):
I don't see why not. Probably we didn't have this earlier because the rlp library was created before u128 was stabilized (1.26.0).

niklasad1 (commented):
Indeed, I agree with Andronik but could you please add a similar decoding test for `u128` to be on the safe side?

niklasad1 (approved):

PR #434, opened 2020-10-23
Add H128 type

drewstone:


primitive-types/src/lib.rs
```diff
@@ -43,6 +43,12 @@ construct_uint! {
 	pub struct U512(8);
 }
 
+construct_fixed_hash! {
+	/// Fixed-size uninterpreted hash type with 20 bytes (160 bits) size.
```
dvdplm, line 47: fix docs please

dvdplm (changes requested):
A motivation would be good, why do we need this?

drewstone:
I'm working on an ETH light-client built off of rainbow bridge and these types are useful.

ordian (approved):

dvdplm (approved):

PR #435, opened 2020-10-26
fixed-hash: Remove allocation in FromStr

e00E:
A similar change could be made in the uint crate.

### DIFF
--- a/fixed-hash/src/hash.rs
+++ b/fixed-hash/src/hash.rs
@@ -588,14 +588,15 @@ macro_rules! impl_rustc_hex_for_fixed_hash {
 			/// - When encountering invalid non hex-digits
 			/// - Upon empty string input or invalid input length in general
 			fn from_str(input: &str) -> $crate::core_::result::Result<$name, $crate::rustc_hex::FromHexError> {
-				#[cfg(not(feature = "std"))]
-				use $crate::alloc_::vec::Vec;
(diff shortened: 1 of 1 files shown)


fixed-hash/src/hash.rs
```diff
@@ -588,14 +588,15 @@ macro_rules! impl_rustc_hex_for_fixed_hash {
...
-				#[cfg(not(feature = "std"))]
-				use $crate::alloc_::vec::Vec;
-				use $crate::rustc_hex::FromHex;
-				let bytes: Vec<u8> = input.from_hex()?;
-				if bytes.len() != Self::len_bytes() {
-					return Err($crate::rustc_hex::FromHexError::InvalidHexLength);
+				let mut iter = $crate::rustc_hex::FromHexIter::new(input);
+				let mut result = Self::zero();
+				for byte in result.as_mut() {
+					*byte = iter.next().ok_or(Self::Err::InvalidHexLength)??;
```
niklasad1, line 594:
```suggestion
					*byte = iter.next().ok_or(Self::Err::InvalidHexLength)?;
```

e00E: The double `?` is needed because the iterator returns `Option<Result<...>>` which this code turns into `Result<Result<...>>`.

niklasad1 (approved):
LGTM

ordian (approved):
Thanks!

PR #436, opened 2020-10-26
Add conversions U256 <-> f64

bh2smith:
Based on an open issue in a dependent repo https://github.com/gnosis/dex-services/issues/1455 which suggests upstreamming the conversion of U256 to and from f64.

This PR introduces two implementations of From for U256 and f64 (essentially mapping between the two). Corresponding unit tests are also included for generic and corner cases.

_Note that_ this code and unit tests have been migrated, almost identically, from two repos

- [here](https://github.com/gnosis/dex-services/blob/3c7c4d44a0ba519992edc9b5deab292f8b23cf6c/pricegraph/src/num.rs#L37-L78) in `gnosis/dex-services`

and

- [here](https://github.com/gnosis/ethcontract-rs/blob/4dd8ab4c173e715a630144592a0d24fbe8d8ed93/src/conv.rs#L5-L43) in `gnosis/ethcontract-rs`

and was initially contributed by @nlordell


ordian:
Hi @bh2smith, thanks for the PR!

I'm not sure we want to have these conversion as `From` impls in `uint`:
* `std::convert::From` is designed for lossless convertion between types, e.g. `f64` impl [`From<u32>`](https://doc.rust-lang.org/std/primitive.f64.html#impl-From%3Cu32%3E), but not `From<u64>`
* the conversion to primitive types in rust is done via https://doc.rust-lang.org/std/convert/trait.FloatToInt.html which is unstable

If this conversions are useful for DEXes (although I'd be extra cautious to use floats in a blockchain space, where determinism is a must), maybe we could add them as a free-standing function, wdyt @dvdplm @niklasad1?

primitive-types/src/lib.rs
```diff
@@ -219,3 +219,45 @@ impl<'a> TryFrom<&'a U512> for U256 {
...
+	/// - `(0, u256::MAX]` => `value as u256`
+	/// - `(u256::MAX, +∞)` => `u256::MAX`
+	fn from(value: f64) -> U256 {
+		if value >= 1.0 {
+			let bits = value.to_bits();
+			// NOTE: Don't consider the sign or check that the subtraction will
+			//   underflow since we already checked that the value is greater
+			//   than 1.0.
+			let exponent = ((bits >> 52) & 0x7ff) - 1023;
+			let mantissa = (bits & 0x0f_ffff_ffff_ffff) | 0x10_0000_0000_0000;
```
ordian, line 240: does this implementation assume little-endianness?

niklasad1:
> I'm not sure we want to have these conversion as From impls in uint

I agree, that it shouldn't be a `From/Into impl`. Free-standing functions or methods such as  `to_f64_lossy` and `from_f64_lossy` would be fine too I guess.

I just think floating point would add additional complexity to this crate and I'm not sure it's worth the effort? However, if enough people wants it, it's fine I guess.

Regarding the `Eq` and `PartialEq` impls, are they fine because they are still represented as a byte array internally?

Robbepop:
> I just think floating point would add additional complexity to this crate and I'm not sure it's worth the effort? However, if enough people wants it, it's fine I guess.

We could put it behind a crate feature.

primitive-types/src/lib.rs
```diff
@@ -110,6 +110,44 @@ impl U256 {
...
+			}
+		} else {
+			0.into()
+		}
+	}
+
+	pub fn to_f64_lossy(self) -> f64 {
+		let (res, factor) = match self {
+			U256([_, _, 0, 0]) => (self, 1.0),
+			U256([_, _, _, 0]) => (self >> 64, 2.0f64.powi(64)),
```
ordian, line 146: it seems powi is not available in `no_std`, hence the failing tests, we could feature gate the `to_f64_lossy` method if there is no alternative

primitive-types/src/lib.rs
```diff
@@ -110,6 +110,44 @@ impl U256 {
...
+	/// - `(0, u256::MAX]` => `value as u256`
+	/// - `(u256::MAX, +∞)` => `u256::MAX`
+	pub fn from_f64_lossy(value: f64) -> U256 {
+		if value >= 1.0 {
+			let bits = value.to_bits();
+			// NOTE: Don't consider the sign or check that the subtraction will
+			//   underflow since we already checked that the value is greater
+			//   than 1.0.
+			let exponent = ((bits >> 52) & 0x7ff) - 1023;
+			let mantissa = (bits & 0x0f_ffff_ffff_ffff) | 0x10_0000_0000_0000;
```
ordian, line 130: could you please confirm it works on big-endian architechtures as well?

primitive-types/src/lib.rs
bh2smith: This is independent of endianness since we are not indexing bytes here.

primitive-types/src/lib.rs
bh2smith: Yes, I have a fix coming!

primitive-types/src/lib.rs
bh2smith: Yes, this is independent of endianness since we are not indexing bytes here. That is, this is a purely uint implementation since above we see that that the bits obtained from `to_bytes` is a u64.

bh2smith:
> We could put it behind a crate feature.

I don't see how including this as a feature would be useful since we do not require any additional libraries.

primitive-types/src/lib.rs
```diff
@@ -110,6 +110,44 @@ impl U256 {
 	pub fn full_mul(self, other: U256) -> U512 {
 		U512(uint_full_mul_reg!(U256, 4, self, other))
 	}
+
+	/// Lossy saturating conversion from a `f64` to a `U256`.
+	///
+	/// The conversion follows roughly the same rules as converting `f64` to other
```
dvdplm, line 116: "roughly"? If there are any differences they should be documented. I'd prefer there were none though.

I'd also like to see some docs/examples about how `f64`s are truncated, e.g.`10.5_f64` –> `10_u256`.

primitive-types/tests/conversion.rs
```diff
@@ -0,0 +1,70 @@
+// Copyright 2020 Parity Technologies
```
dvdplm, line 1: This file should probably be named `fp_conversions.rs` or something that conveys the content better.

dvdplm (commented):
I'm ok to include this, but let's use a feature gate.

dvdplm:
> I don't see how including this as a feature would be useful since we do not require any additional libraries.

Fair point, but I think it's more of a communication thing than purely technical. I think this is a niche feature (with sharp edges, like @ordian mentions) and feature gating it puts up that little extra barrier to users. They need to really want this to get it, sort of?

primitive-types/src/lib.rs
ordian: > Yes, this is independent of endianness since we are not indexing bytes here. That is, this is a purely uint implementation since above we see that that the bits obtained from `to_bytes` is a u64.

So the bytes within this `u64` can be either in little-endian form or big-endian depending on our CPU.
My understanding is that `bits & 0x0f_ffff_ffff_ffff` assumes the little-endianness, please convince me that I'm wrong and this is specified by IEEE-754.
From https://en.wikipedia.org/wiki/Double-precision_floating-point_format#Endianness
> However, on modern standard computers (i.e., implementing IEEE 754), one may in practice safely assume that the endianness is the same for floating-point numbers as for integers

To be on the safer side, I'd suggest using `f64::to_be_bytes`.

primitive-types/src/lib.rs
bh2smith: Yes, cool idea. Would it suffice to have a unit test like so:

```rs
#[test]
fn f64_to_u256_truncation_examples() {
	assert_eq!(U256::from_f64_lossy(10.5), 10.into());
}
```

with some examples? Do you have any other examples in mind? I ask because most of these corner cases are already covered in the tests.

dvdplm: I agree that the unit tests seem to cover it all, I was referring to the docs. If you want to also provide examples that's even better.

So yeah, avoid vague language like "roughly" (or expand on it) and explain to users how their floats are cast, "The conversion behaves like [`trunc`](https://doc.rust-lang.org/std/primitive.f64.html#method.trunc)." or something along those lines?

primitive-types/src/lib.rs
bh2smith: According to the docs https://doc.rust-lang.org/stable/std/primitive.f64.html#method.from_bits

> It turns out this is incredibly portable, for two reasons:
> 
> - Floats and Ints have the same endianness on all supported platforms.
> - IEEE-754 very precisely specifies the bit layout of floats.

ordian: Ok, I looked at the generated code for a big-endian CPU and it seems to work https://godbolt.org/z/jP6rzK.
Seems like this assumption was wrong:
> My understanding is that bits & 0x0f_ffff_ffff_ffff assumes the little-endianness

Thanks for bearing with me.

primitive-types/tests/fp_conversion.rs
```diff
@@ -0,0 +1,78 @@
...
+//
+// Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
+// http://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+// <LICENSE-MIT or http://opensource.org/licenses/MIT>, at your
+// option. This file may not be copied, modified, or distributed
+// except according to those terms.
+
+//! Tests for to and from f64 lossy for U256 primitive-type.
+
+#![cfg(feature = "fp-conversion")]
```
ordian, line 11: see the bottom of the `primitive-types/Cargo.toml` for how it should be done
```toml
[[test]]
name = "fp_conversion"
required-features = ["fp-conversion"]
```

ordian (approved):

primitive-types/tests/fp_conversion.rs
bh2smith: Thanks, made the change here!

bh2smith:
All the suggested changes have been made! I hope it is good shape for you now @dvdplm!

dvdplm (approved):

dvdplm:
@bh2smith Looks good. Thank you so for your work on this. :)

bh2smith:
This is great! Thanks for including this change. Is there a planned release for the next version so that we can update our repos with this?

fleupold:
@dvdplm would it possible to do a minor version release with this change? Depending on it via the git repo/commit requires a huge download.

ordian:
@fleupold sorry for the delay, opened #440.

dvdplm:
@fleupold @bh2smith `0.7.3` is published.

fleupold:
<a href="https://gitme.me/image?url=https%3A%2F%2Fmedia1.giphy.com%2Fmedia%2FKJ1f5iTl4Oo7u%2Fgiphy-downsized-medium.gif&token=thanks" data-gitmeme-token="thanks"></a>

PR #462, opened 2020-12-01
rlp: Fix buffer indexing

vorot93:
Users can provide their own buffer now - we must assume that buffers start with arbitrary length.

### DIFF
--- a/rlp/src/stream.rs
+++ b/rlp/src/stream.rs
@@ -29,6 +29,7 @@ impl ListInfo {
 /// Appendable rlp encoder.
 pub struct RlpStream {
 	unfinished_lists: Vec<ListInfo>,
+	start_pos: usize,
 	buffer: BytesMut,
 	finished_list: bool,
 }
@@ -52,7 +53,7 @@ impl RlpStream {
 
 	/// Initializes instance of empty `Stream`.
 	pub fn new_with_buffer(buffer: BytesMut) -> Self {
-		RlpStream { unfinished_lists: Vec::with_capacity(16), buffer, finished_list: false }
+		RlpStream { unfinished_lists: Vec::with_capacity(16), start_pos: buffer.len(), buffer, finished_list: false }
 	}
 
 	/// Initializes the `Stream` as a list.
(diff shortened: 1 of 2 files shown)


ordian (approved):
Could you add a test for `clear` as well?

rlp/src/stream.rs
```diff
@@ -359,11 +369,11 @@ impl<'a> BasicEncoder<'a> {
 		// 1 byte was already reserved for payload earlier
 		match len {
 			0..=55 => {
-				self.buffer[pos - 1] = 0xc0u8 + len as u8;
+				self.buffer[self.start_pos + pos - 1] = 0xc0u8 + len as u8;
```
dvdplm, line 372: If `pos` is ever zero it looks like nasty things could happen (was true before too); can you double-check it cannot be called with `pos` == 0?

rlp/src/stream.rs
```diff
@@ -403,9 +413,9 @@ impl<'a> BasicEncoder<'a> {
 			// (prefix + length of length), followed by the length, followd by the string
 			len => {
 				self.buffer.put_u8(0);
-				let position = self.buffer.len();
+				let position = self.total_written();
 				let inserted_bytes = self.insert_size(len, position);
-				self.buffer[position - 1] = 0xb7 + inserted_bytes;
+				self.buffer[self.start_pos + position - 1] = 0xb7 + inserted_bytes;
```
dvdplm, line 418: Same concern here, if both `start_pos` and `position` can be `0` (and `self.total_written()` starts at `0` right?) then we'd get an out of bounds panic right?

rlp/src/stream.rs
vorot93: I cannot be sure without auditing all possible encoding paths in this codebase, but this is orthogonal to this PR as it _increases_ the count.

I suggest we merge as is, and add this concern to a round of fuzzing which this library so badly needs anyway.

rlp/src/stream.rs
vorot93: see above

rlp/src/stream.rs
niklasad1: Yepp, agree unrelated to this PR but we should open a ticket for it and investigate it further.

niklasad1 (approved):

PR #465, opened 2020-12-02
rlp: store test bytes in hex literals

vorot93:
A minor change to tests, improves readability for users who read our tests as a reference for how RLP works.


dvdplm (approved):
I'm not sure I agree that this is much of an improvement to readability, but if you're confident this is easier for newcomers to read, why not?

ordian (approved):

PR #466, opened 2020-12-03
Make RLP optional in several crates

vorot93:
For symmetry with serde and SCALE impls. Also, RLP is on its way out of Ethereum, so for future-proofing as well.


ordian (approved):
I'm pretty sure this was intended, but OTOH there is no reason to disallow using `ethereum-types` without `rlp` impl.

niklasad1 (approved):
:+1:

PR #487, opened 2020-12-27
fix: make from_str parse 0x-prefixed strings

gakonst:
As title

Travis' error seems unrelated
image

### DIFF
--- a/fixed-hash/src/hash.rs
+++ b/fixed-hash/src/hash.rs
@@ -588,7 +588,11 @@ macro_rules! impl_rustc_hex_for_fixed_hash {
 			/// - When encountering invalid non hex-digits
 			/// - Upon empty string input or invalid input length in general
 			fn from_str(input: &str) -> $crate::core_::result::Result<$name, $crate::rustc_hex::FromHexError> {
-				let input = if input.starts_with("0x") { &input[2..] } else { input };
+                let input = if input.starts_with("0x") {
+                    &input[2..]
+                } else {
+                    input
+                };
 				let mut iter = $crate::rustc_hex::FromHexIter::new(input);
 				let mut result = Self::zero();
 				for byte in result.as_mut() {
--- a/fixed-hash/src/hash.rs
+++ b/fixed-hash/src/hash.rs
@@ -588,11 +588,11 @@ macro_rules! impl_rustc_hex_for_fixed_hash {
 			/// - When encountering invalid non hex-digits
 			/// - Upon empty string input or invalid input length in general
 			fn from_str(input: &str) -> $crate::core_::result::Result<$name, $crate::rustc_hex::FromHexError> {
-                let input = if input.starts_with("0x") {
-                    &input[2..]
-                } else {
-                    input
-                };
+				let input = if input.starts_with("0x") {
+					&input[2..]
+				} else {
+					input
+				};
 				let mut iter = $crate::rustc_hex::FromHexIter::new(input);
 				let mut result = Self::zero();
 				for byte in result.as_mut() {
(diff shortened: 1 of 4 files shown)


uint/src/uint.rs
```diff
@@ -1688,6 +1688,11 @@ macro_rules! construct_uint {
 			type Err = $crate::FromHexError;
 
 			fn from_str(value: &str) -> $crate::core_::result::Result<$name, Self::Err> {
+                let value = if value.starts_with("0x") {
```
dvdplm, line 1691: Indentation looks off here.

fixed-hash/src/hash.rs
```diff
@@ -588,6 +588,7 @@ macro_rules! impl_rustc_hex_for_fixed_hash {
 			/// - When encountering invalid non hex-digits
 			/// - Upon empty string input or invalid input length in general
 			fn from_str(input: &str) -> $crate::core_::result::Result<$name, $crate::rustc_hex::FromHexError> {
+				let input = if input.starts_with("0x") { &input[2..] } else { input };
```
dvdplm, line 591: Prefer multiline here.

uint/tests/uint_tests.rs
```diff
@@ -132,7 +132,9 @@ fn uint256_from() {
 	);
 
 	// test initializtion from string
+	let sa2 = U256::from_str("0x0a").unwrap();
 	let sa = U256::from_str("0a").unwrap();
+	assert_eq!(sa2, sa);
```
dvdplm, line 137:
```suggestion
	let sa = U256::from_str("0a").unwrap();
	let sa2 = U256::from_str("0x0a").unwrap();
	assert_eq!(sa2, sa);
```

dvdplm (changes requested):

uint/src/uint.rs
gakonst: Seems like a github bug. It doesn't show in the Full review tab, and cargo fmt doesn't catch it

image

image

dvdplm: It looks like you're using spaces (all crates here use tabs, `\t`) and that's why it renders weird on github. Can you convert to tabs please?

gakonst:
@dvdplm I've addressed your changes, would appreciate another look - thanks!

dvdplm (approved):

gakonst:
@niklasad1 @ordian would be great if you could approve this PR

ordian (approved):

dvdplm:
@gakonst We need a `cargo fmt` I'm afraid.

gakonst:
@dvdplm I did that initially and you said you prefer multiline :) Addressed in latest commit. Thanks!

fixed-hash/src/hash.rs
ordian, line 591: alternatively we could use
```suggestion
				let input = input.strip_prefix("0x").unwrap_or(input);
```

dvdplm:
> @dvdplm I did that initially and you said you prefer multiline :) Addressed in latest commit. Thanks!

I know, my bad. Apologies. :(

gakonst:
Switched to use `strip_prefix`, good rec! PR should be good to merge now.

uint/src/uint.rs
```diff
@@ -1688,6 +1688,7 @@ macro_rules! construct_uint {
 			type Err = $crate::FromHexError;
 
 			fn from_str(value: &str) -> $crate::core_::result::Result<$name, Self::Err> {
+                let value = value.strip_prefix("0x").unwrap_or(value);
```
ordian, line 1691: illegal spaces!

PR #497, opened 2021-01-05
Update rand to 0.8

vorot93:


vorot93:
Okay, looks like I had a git glitch, so didn't see that rand has already been updated in master.

PR #498, opened 2021-01-06
bump fs-swap

xlc:
To include https://github.com/debris/fs-swap/pull/10 so it builds on Apple M1


dvdplm (approved):
Thank you! :)

ordian (approved):
We're not going to make a new release, because it is a minor bump and can be done in `Cargo.lock` of your application.

PR #579, opened 2021-08-05
kvdb api uses now template argument `P: AsRef<Path>` instead of `&str`

debris:
kvdb api uses now template argument `P: AsRef<Path>` instead of `&str`

why?

it reduces the clutter caused by constant conversion of `&Path` to `&str` in crates using `kvdb-rocksdb` (and tests)

also, modified the changelog and the version to 0.14

### DIFF
--- a/kvdb-rocksdb/src/lib.rs
+++ b/kvdb-rocksdb/src/lib.rs
@@ -34,8 +34,6 @@ use regex::Regex;
 #[cfg(target_os = "linux")]
 use std::fs::File;
 #[cfg(target_os = "linux")]
-use std::path::PathBuf;
-#[cfg(target_os = "linux")]
 use std::process::Command;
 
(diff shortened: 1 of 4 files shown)


kvdb-rocksdb/src/lib.rs
```diff
@@ -9,7 +9,14 @@
...
+	cmp,
+	collections::HashMap,
+	convert::identity,
+	error, fs, io, mem,
+	path::{Path, PathBuf},
+	result,
+};
 
 use parity_util_mem::MallocSizeOf;
 use parking_lot::RwLock;
```
niklasad1, line 22: Couldn't comment but remove the import on line 30 below and it should compile!

arkpar (approved):

niklasad1 (approved):
LGTM

PR #602, opened 2021-11-27
Stop calling `state.finish()` for fixed hashes

jakelee8:
Calling [`state.finish();`][Hasher::finish] is a no-op. At call site, `state` is read-only and the return value is ignored. The call and may be removed.

[Hasher::finish]: https://doc.rust-lang.org/std/hash/trait.Hasher.html#tymethod.finish


ordian (approved):
Thanks!

bkchr (approved):

PR #626, opened 2022-02-07
Update scale-info to v2

wigy-opensource-developer:
We should get to a release of primitive-types that could be included in a Substrate version that depends on Scale v3. Yeah :smile:

### DIFF
--- a/primitive-types/Cargo.toml
+++ b/primitive-types/Cargo.toml
@@ -15,7 +15,8 @@ impl-serde = { version = "0.3.1", path = "impls/serde", default-features = false
 impl-codec = { version = "0.6.0", path = "impls/codec", default-features = false, optional = true }
 impl-num-traits = { version = "0.1.0", path = "impls/num-traits", default-features = false, optional = true }
 impl-rlp = { version = "0.3", path = "impls/rlp", default-features = false, optional = true }
-scale-info-crate = { package = "scale-info", version = ">=0.9, <2", features = ["derive"], default-features = false, optional = true }
+# See https://github.com/paritytech/parity-common/pull/556#issuecomment-872127496 why this is here:
+scale-info-crate = { package = "scale-info", version = "2", features = ["derive"], default-features = false, optional = true }
 
 [features]
 default = ["std"]


primitive-types/Cargo.toml
```diff
@@ -15,7 +15,8 @@ impl-serde = { version = "0.3.1", path = "impls/serde", default-features = false
 impl-codec = { version = "0.6.0", path = "impls/codec", default-features = false, optional = true }
 impl-num-traits = { version = "0.1.0", path = "impls/num-traits", default-features = false, optional = true }
 impl-rlp = { version = "0.3", path = "impls/rlp", default-features = false, optional = true }
-scale-info-crate = { package = "scale-info", version = ">=0.9, <2", features = ["derive"], default-features = false, optional = true }
+# See https://github.com/paritytech/parity-common/pull/556#issuecomment-872127496 why this is here:
+scale-info-crate = { package = "scale-info", version = "2", features = ["derive"], default-features = false, optional = true }
```
dvdplm, line 19: @wigy-opensource-developer shouldn't this be `version = "3"`?

wigy-opensource-developer: parity-scale-codec is now v3, and scale-info v2 depends on it. Funsies.

ordian: I think you're conflating parity-scale-codec (v3) with scale-info (v2)

dvdplm:
@wigy-opensource-developer As @ordian points out, we'll need a major bump of `primitive-types`, which will cascade into something akin to [this horror](https://github.com/paritytech/parity-common/pull/623). :/

ordian:
writing comments on Github at the same time is fun :)

primitive-types/Cargo.toml
```diff
@@ -15,7 +15,8 @@ impl-serde = { version = "0.3.1", path = "impls/serde", default-features = false
 impl-codec = { version = "0.6.0", path = "impls/codec", default-features = false, optional = true }
 impl-num-traits = { version = "0.1.0", path = "impls/num-traits", default-features = false, optional = true }
 impl-rlp = { version = "0.3", path = "impls/rlp", default-features = false, optional = true }
-scale-info-crate = { package = "scale-info", version = ">=0.9, <2", features = ["derive"], default-features = false, optional = true }
+# See https://github.com/paritytech/parity-common/pull/556#issuecomment-872127496 why this is here:
```
ordian, line 18: this link doesn't explain why this is here, which is I think self-explanatory (it's a feature)

wigy-opensource-developer:
Sorry, it seems like we all have different scopes and some crates in the big-big DAG of cumulus dependencies created another version ripple here. Maybe we need a better way to navigate that big-big DAG.

surprised

primitive-types/Cargo.toml
ordian, line 19: Ok, pull this YOLO trick once again
```suggestion
scale-info-crate = { package = "scale-info", version = ">=0.9, <4", features = ["derive"], default-features = false, optional = true }
```

wigy-opensource-developer: `<3` you meant? Also, isn't it better to exclude 0.9 and 1.0.0 now?

dvdplm: With this we'd only need a minor release of `primitive-types` yes?

ordian: Yes, I meant <3 and yes, this would be a patch release. And we don't need to release anything else. Only update substrate/polkadot/cumulus

wigy-opensource-developer: Okay, I get it. Although MSRV and edition 2021 is already in the Cargo.toml which technically makes it a major release, right?

ordian: Could you bump the version in this PR and update the change as well please?

ordian: Ah, wait, scale-info also used in ethereum-types. I'm afraid, there's no getting away with breaking release then (sorry).

ordian:
closing in favor of #627

wigy-opensource-developer:
Wow. Thanks a lot. I had to be AFK for 3 hours and you took the bullet for me.

Issue #638, opened 2022-03-30
uint: make is_zero const


webmaster128:
Heyho! It would be very nice if it is possible to make `is_zero` a `const` function. Looking at the implementation, this seems simple to do.

PR #639, opened 2022-03-30
`uint`: Make `is_zero` implementation const

ueco-jb:
closes https://github.com/paritytech/parity-common/issues/638

`for` loops are not allowed in const functions, so I went with simple `while` loop approach.

### DIFF
--- a/uint/src/uint.rs
+++ b/uint/src/uint.rs
@@ -673,9 +673,10 @@ macro_rules! construct_uint {
 
 			/// Whether this is zero.
 			#[inline]
-			pub fn is_zero(&self) -> bool {
+			pub const fn is_zero(&self) -> bool {
 				let &$name(ref arr) = self;
-				for i in 0..$n_words { if arr[i] != 0 { return false; } }
+				let mut i = 0;
+				while i < $n_words { if arr[i] != 0 { return false; } else { i += 1; } }
 				return true;
 			}
 


niklasad1 (approved):
LGTM, thank you

ueco-jb:
@niklasad1 I need someone else to approve here, because 2 reviews are required.
Could you ask someone from group?

arkpar (approved):

uint/src/uint.rs
```diff
@@ -673,9 +673,10 @@ macro_rules! construct_uint {
 
 			/// Whether this is zero.
 			#[inline]
-			pub fn is_zero(&self) -> bool {
+			pub const fn is_zero(&self) -> bool {
 				let &$name(ref arr) = self;
-				for i in 0..$n_words { if arr[i] != 0 { return false; } }
+				let mut i = 0;
+				while i < $n_words { if arr[i] != 0 { return false; } else { i += 1; } }
```
kianenigma, line 679: so while is now stable to loop over in const context but `for` not yet? interesting, seemed like all of them are behind `const_looping`

ueco-jb: Yeah, `for` loop calls `iter.next()` which is not a `const fn`.
Future RFCs (like https://github.com/rust-lang/rfcs/pull/2237) might lift that restriction.

PR #682, opened 2022-09-28
ethereum-types: add if_ethbloom conditional macro

vorot93:
In the spirit of serde's [serde_if_integer128](https://docs.rs/serde/1/serde/macro.serde_if_integer128.html).


ordian (approved):
I think this is sensible.

PR #713, opened 2023-01-26
bounded_vec: Avoid unnecessary allocation when decoding

mrcnski:
Needed for https://github.com/paritytech/polkadot/pull/6603. See https://github.com/paritytech/polkadot/pull/6603#discussion_r1083461147:

> You want to ensure that you don't even start try decoding/allocating when the length of the vector is more than the
> allowed maximum.

> You first decode length of the vector and then early reject if that is too long.

### DIFF
--- a/bounded-collections/src/bounded_btree_map.rs
+++ b/bounded-collections/src/bounded_btree_map.rs
@@ -400,6 +400,7 @@ mod test {
 	use super::*;
 	use crate::ConstU32;
 	use alloc::{vec, vec::Vec};
+	use codec::CompactLen;
 
 	fn map_from_keys<K>(keys: &[K]) -> BTreeMap<K, ()>
 	where
@@ -414,6 +415,14 @@ mod test {
 		S: Get<u32>,
 	{
 		map_from_keys(keys).try_into().unwrap()
+	}
+
+	#[test]
+	fn encoding_same_as_unbounded_map() {
+		let b = boundedmap_from_keys::<u32, ConstU32<7>>(&[1, 2, 3, 4, 5, 6]);
+		let m = map_from_keys(&[1, 2, 3, 4, 5, 6]);
+
+		assert_eq!(b.encode(), m.encode());
 	}
 
 	#[test]
@@ -464,6 +473,16 @@ mod test {
 			BoundedBTreeMap::<u32, u32, ConstU32<4>>::decode(&mut &v.encode()[..]),
 			Err("BoundedBTreeMap exceeds its limit".into()),
 		);
+	}
+
+	#[test]
+	fn dont_consume_more_data_than_bounded_len() {
+		let m = map_from_keys(&[1, 2, 3, 4, 5, 6]);
+		let data = m.encode();
+		let data_input = &mut &data[..];
+
+		BoundedBTreeMap::<u32, u32, ConstU32<4>>::decode(data_input).unwrap_err();
+		assert_eq!(data_input.len(), data.len() - Compact::<u32>::compact_len(&(data.len() as u32)));
 	}
 
 	#[test]
--- a/bounded-collections/src/bounded_btree_map.rs
+++ b/bounded-collections/src/bounded_btree_map.rs
@@ -19,7 +19,7 @@
 
 use crate::{Get, TryCollect};
 use alloc::collections::BTreeMap;
-use codec::{Decode, Encode, MaxEncodedLen};
+use codec::{Compact, Decode, Encode, MaxEncodedLen};
 use core::{borrow::Borrow, marker::PhantomData, ops::Deref};
 
 /// A bounded map based on a B-Tree.
@@ -29,6 +29,7 @@ use core::{borrow::Borrow, marker::PhantomData, ops::Deref};
 ///
 /// Unlike a standard `BTreeMap`, there is an enforced upper limit to the number of items in the
 /// map. All internal operations ensure this bound is respected.
+#[cfg_attr(feature = "std", derive(Hash))]
 #[derive(Encode, scale_info::TypeInfo)]
 #[scale_info(skip_type_params(S))]
 pub struct BoundedBTreeMap<K, V, S>(BTreeMap<K, V>, PhantomData<S>);
@@ -40,10 +41,15 @@ where
 	S: Get<u32>,
 {
 	fn decode<I: codec::Input>(input: &mut I) -> Result<Self, codec::Error> {
-		let inner = BTreeMap::<K, V>::decode(input)?;
-		if inner.len() > S::get() as usize {
+		// Same as the underlying implementation for `Decode` on `BTreeMap`, except we fail early if
+		// the len is too big.
+		let len: u32 = <Compact<u32>>::decode(input)?.into();
+		if len > S::get() {
 			return Err("BoundedBTreeMap exceeds its limit".into())
 		}
+		input.descend_ref()?;
+		let inner = Result::from_iter((0..len).map(|_| Decode::decode(input)))?;
+		input.ascend_ref();
 		Ok(Self(inner, PhantomData))
 	}
 
--- a/bounded-collections/src/bounded_btree_set.rs
+++ b/bounded-collections/src/bounded_btree_set.rs
@@ -330,6 +330,7 @@ mod test {
 	use super::*;
 	use crate::ConstU32;
 	use alloc::{vec, vec::Vec};
+	use codec::CompactLen;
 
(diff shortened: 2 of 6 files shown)


bounded-collections/src/bounded_vec.rs
```diff
@@ -289,10 +289,11 @@ impl<'a, T, S: Get<u32>> BoundedSlice<'a, T, S> {
 
 impl<T: Decode, S: Get<u32>> Decode for BoundedVec<T, S> {
 	fn decode<I: codec::Input>(input: &mut I) -> Result<Self, codec::Error> {
-		let inner = Vec::<T>::decode(input)?;
-		if inner.len() > S::get() as usize {
+		let len: u32 = <Compact<u32>>::decode(input)?.into();
+		if len > S::get() {
```
ggwpez, line 293: `BoundedBTreeMap` and `BoundedBTreeSet` also suffer from this.  
Could you please also fix it there?

bkchr (approved):
% what @ggwpez said.

KiChjang (approved):
Also probably requires a version bump for substrate to pick up the changes.

mrcnski:
> Also probably requires a version bump for substrate to pick up the changes.

Should I do that in this PR?

Also, I'll try to get to `BoundedBTreeMap` and `BoundedBTreeSet` soon, haven't had a chance yet.

ggwpez (approved):
>Should I do that in this PR?

You can also do it in another PR, just thought it would fit in here :smile:

bkchr:
As the crate is currently lacking test, I would be happy if this pr could start with this ;) We could ensure that we don't consume more data than the compact length.

mrcnski:
@bkchr There are tests, including for that case -- see `too_big_vec_fail_to_decode` and `too_big_fail_to_decode`. Or do you mean we should check whether unnecessary memory is consumed (allocated)? I wasn't sure how to do that, or if such a test was necessary.

mrcnski:
This PR is already approved, but would like a quick re-approval from someone (to sanity check the last two commits). :)

ggwpez:
>I wasn't sure how to do that, or if such a test was necessary.

Dump idea: Just use `u32` as Bound and `1MiB` as value, then try to decode that. Should fail to allocate if the implementation is wrong.

bkchr:
> I wasn't sure how to do that, or if such a test was necessary.

```
let data = something.encode();
let data_input = &mut &data[..];

Something::decode(data_input).unwrap_err();
assert_eq!(data_input.len(), data.len() - Compact::<u32>(data.len() as u32).compact_len());
```

This should do it. Also some easy tests ensuring that `Vec` etc encodings are the same as the bounded impls.

> There are tests, including for that case -- see `too_big_vec_fail_to_decode` and `too_big_fail_to_decode`

Sorry :see_no_evil:

bounded-collections/src/bounded_btree_map.rs
```diff
@@ -40,10 +41,15 @@ where
...
-		let inner = BTreeMap::<K, V>::decode(input)?;
-		if inner.len() > S::get() as usize {
+		// Same as the underlying implementation for `Decode` on `BTreeMap`, except we fail early if
+		// the len is too big.
+		let len: u32 = <Compact<u32>>::decode(input)?.into();
+		if len > S::get() {
 			return Err("BoundedBTreeMap exceeds its limit".into())
 		}
+		input.descend_ref()?;
+		let inner = Result::from_iter((0..len).map(|_| Decode::decode(input)))?;
```
ggwpez, line 51: No idea if rust handles this in the most efficient way, otherwise we should maybe pre-allocate as optimization.

bounded-collections/src/bounded_btree_set.rs
```diff
@@ -39,10 +40,15 @@ where
...
 	fn decode<I: codec::Input>(input: &mut I) -> Result<Self, codec::Error> {
-		let inner = BTreeSet::<T>::decode(input)?;
-		if inner.len() > S::get() as usize {
+		// Same as the underlying implementation for `Decode` on `BTreeSet`, except we fail early if
+		// the len is too big.
+		let len: u32 = <Compact<u32>>::decode(input)?.into();
+		if len > S::get() {
 			return Err("BoundedBTreeSet exceeds its limit".into())
 		}
+		input.descend_ref()?;
```
ggwpez, line 49: Have not seen these functions yet, but LGTM

ggwpez (approved):

bounded-collections/src/bounded_btree_map.rs
bkchr: I mean this is now discusable with the bound, but for the normal `BTreeMap` we don't have done this as someone could send you a fake encoded `BTreeMap` that pretends to be `u32::Max` in size and then maybe leading to some OOM.

mrcnski: Interesting, would have to compile with optimizations and see. At any rate, I don't want to deviate too much from the standard `Decode` for the unbounded types, at least in this PR.

bkchr: Should be fine to merge as is.

bounded-collections/src/bounded_btree_map.rs
```diff
@@ -29,6 +29,7 @@ use core::{borrow::Borrow, marker::PhantomData, ops::Deref};
 ///
 /// Unlike a standard `BTreeMap`, there is an enforced upper limit to the number of items in the
 /// map. All internal operations ensure this bound is respected.
+#[cfg_attr(feature = "std", derive(Hash))]
```
bkchr, line 32:
```suggestion
#[cfg_attr(feature = "std", derive(Hash, serde::Serialize, serde::Deserialize), serde(transparent))]
```

bounded-collections/src/bounded_btree_set.rs
```diff
@@ -29,6 +29,7 @@ use core::{borrow::Borrow, marker::PhantomData, ops::Deref};
 ///
 /// Unlike a standard `BTreeSet`, there is an enforced upper limit to the number of items in the
 /// set. All internal operations ensure this bound is respected.
+#[cfg_attr(feature = "std", derive(Hash))]
```
bkchr, line 32:
```suggestion
#[cfg_attr(feature = "std", derive(Hash, serde::Serialize, serde::Deserialize), serde(transparent))]
```

bounded-collections/src/bounded_vec.rs
```diff
@@ -40,7 +40,7 @@ use serde::{
 ///
 /// As the name suggests, the length of the queue is always bounded. All internal operations ensure
 /// this bound is respected.
-#[cfg_attr(feature = "std", derive(Serialize), serde(transparent))]
+#[cfg_attr(feature = "std", derive(Hash, Serialize), serde(transparent))]
```
bkchr, line 43:
```suggestion
#[cfg_attr(feature = "std", derive(Hash, serde::Serialize, serde::Deserialize), serde(transparent))]
```

bkchr (approved):
Ty! But let us use the same derives everywhere.

bounded-collections/src/bounded_btree_map.rs
mrcnski: I'm not sure about deriving these; `Deserialize` would need a custom `impl` to add bounds checks, like we do now with `BoundedVec`. And the custom `impl` looks annoying to implement, so I'm thinking we don't add it until there's actually a need for it.

bounded-collections/src/bounded_btree_map.rs
KiChjang: Is this just another way of calling `collect`?

```suggestion
		let inner = (0..len).map(|_| Decode::decode(input)).collect::<Result<Vec<_>, codec::Error>>()?;
```

mrcnski: Yes, I think it is. http://doc.rust-lang.org/1.66.1/std/result/enum.Result.html#method.from_iter

Currently this impl is the same as the unbounded version, apart from the length check. I think it should be kept the same unless there's a good reason not to.

bounded-collections/src/bounded_btree_map.rs
mrcnski: @bkchr I'm going to go ahead and merge -- these derives can be a follow-up if we really want them.

mrcnski:
@KiChjang Looks like I'm not able to publish `0.1.4`, you're the only listed owner. Possible to add [paritytech/Core devs](https://crates.io/teams/github:paritytech:core-devs) as an owner?

KiChjang:
I've added the core-devs org as the owner for the crate on crates.io.

PR #721, opened 2023-02-13
bounded-collections: Fixes `Hash` impl

mrcnski:
Previously it could not be used in practice, because the size bound was required to also implement `Hash`.

See https://github.com/paritytech/polkadot/pull/6603#discussion_r1098914268.

Also bumps the patch version.


bkchr (approved):

mrcnski:
Thanks @lexnv, that was fast. 🚀

PR #722, opened 2023-02-16
Introduces ConstGet<T>

gupnik:
Issue: https://github.com/paritytech/substrate/issues/9967

### DIFF
--- a/bounded-collections/src/lib.rs
+++ b/bounded-collections/src/lib.rs
@@ -208,7 +208,7 @@ macro_rules! parameter_types {
 			}
 		}
 
-		impl<I: From<$type>> $crate::Get<I> for $name {
+		impl<I: From<$type>> $crate::ConstGet<I> for $name {
 			fn get() -> I {
 				I::from(Self::get())
 			}
--- a/bounded-collections/src/lib.rs
+++ b/bounded-collections/src/lib.rs
@@ -41,15 +41,33 @@ pub trait Get<T> {
 	fn get() -> T;
 }
 
-impl<T: Default> Get<T> for () {
+/// A trait for querying a single value from a type.
+///
+/// The value should be a constant.
+pub trait ConstGet<T> {
+	/// Return the constant value.
+	fn get() -> T;
+}
+
+/// Implement Get for any type that implements ConstGet.
+impl<T, U> Get<U> for T
+where
+	T: ConstGet<U>,
+{
+	fn get() -> U {
+		<T as ConstGet<U>>::get()
+	}
+}
+
+impl<T: Default> ConstGet<T> for () {
 	fn get() -> T {
 		T::default()
 	}
 }
 
-/// Implement Get by returning Default for any type that implements Default.
+/// Implement ConstGet by returning Default for any type that implements Default.
 pub struct GetDefault;
-impl<T: Default> Get<T> for GetDefault {
+impl<T: Default> ConstGet<T> for GetDefault {
 	fn get() -> T {
 		T::default()
 	}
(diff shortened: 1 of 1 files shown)


shawntabrizi (approved):

bounded-collections/src/lib.rs
```diff
@@ -41,15 +41,33 @@ pub trait Get<T> {
...
-impl<T: Default> Get<T> for () {
+/// A trait for querying a single value from a type.
+///
+/// The value should be a constant.
+pub trait ConstGet<T> {
+	/// Return the constant value.
+	fn get() -> T;
+}
+
+/// Implement Get for any type that implements ConstGet.
```
kianenigma, line 52:
```suggestion
// Implement Get for any type that implements ConstGet.
```

impl blocks don't have rust-docs I think and should be documented with just `//`.

bkchr: They can have docs. They also appear in the docs.

bounded-collections/src/lib.rs
```diff
@@ -41,15 +41,33 @@ pub trait Get<T> {
...
+impl<T, U> Get<U> for T
```
KiChjang, line 53: This implementation can be omitted if we just make `Get<T>` a supertrait of `ConstGet<T>`.

bounded-collections/src/lib.rs
gupnik: Used supertrait as suggested by @KiChjang.

bounded-collections/src/lib.rs
gupnik: Done, thanks!

bounded-collections/src/lib.rs
```diff
@@ -76,6 +97,16 @@ macro_rules! impl_const_get {
...
 			}
 		}
+		impl<const T: $t> ConstGet<$t> for $name<T> {
+			fn get() -> $t {
+				T
+			}
+		}
+		impl<const T: $t> ConstGet<Option<$t>> for $name<T> {
+			fn get() -> Option<$t> {
+				Some(T)
```
KiChjang, line 107: Why do we need this implementation?

gupnik: It actually implements `ConstGet` for all `const` types.

KiChjang: I meant why do we need an implementation of `ConstGet<Option<T>>` on the `Const*` types.

gupnik: Ah, I was following a similar paradigm as `Get<Option<T>>` that has already been implemented for const types.

KiChjang (approved):

bounded-collections/src/lib.rs
```diff
@@ -41,20 +41,41 @@ pub trait Get<T> {
...
 }
 
-/// Implement Get by returning Default for any type that implements Default.
+impl<T: Default> ConstGet<T> for () {
+	fn get() -> T {
+		T::default()
+	}
+}
+
+/// Implement ConstGet by returning Default for any type that implements Default.
```
ggwpez, line 64: This comment looks out of place here.

bounded-collections/src/lib.rs
```diff
@@ -41,20 +41,41 @@ pub trait Get<T> {
 	fn get() -> T;
 }
 
+/// A trait for querying a single value from a type.
+///
+/// The value should be a constant.
```
ggwpez, line 46: Do i under stand correctly that the only difference to the normal `Get` trait is that this **MUST** be implemented by a const?  
I think that should be highlighted a bit better as to how it differs from `Get`.

ggwpez (commented):
Test?

gupnik:
@ggwpez I think @kianenigma mentioned that this doesn't add a lot of value as `Const` still doesn't enforce that the provided value is indeed a const and is more of a syntatic sugar at this point. So, will close this one.

PR #734, opened 2023-03-28
Fixed hash type for BLS double public key type and BLS signature along with its DLEQ proof pair. 

drskalman:
Adds H896 and H1152 fixed hashed types. 
H896 = (48 Bytes for BLS signature + 32*2 for the two scalers for the DLEQ proof).
H1152 = (48 Bytes for G1 Public key + 96 Bytes for G2 Public Key).


primitive-types/src/lib.rs
```diff
@@ -77,6 +77,16 @@ construct_fixed_hash! {
 	#[cfg_attr(feature = "scale-info", derive(TypeInfo))]
 	pub struct H768(96);
 }
+construct_fixed_hash! {
+	/// Fixed-size uninterpreted hash type with 160 bytes (896 bits) size.
```
davxy, line 81:
```suggestion
	/// Fixed-size uninterpreted hash type with 112 bytes (896 bits) size.
```

davxy:
@drskalman please first read this : https://github.com/paritytech/substrate/pull/13618#discussion_r1151085564

davxy:
Closed because we're not gonna need it

PR #385, opened 2020-05-01
primitive-types: add no_std support for serde feature

xu-cheng:
This adds no_std support to primitive-types with serde.

Due to https://github.com/rust-lang/cargo/issues/3494,
a separate new feature `serde_no_std` is created.

PR #745, opened 2023-04-28
`serde` feature added

michalkucharczyk:
Support for `serde` derivations in no_std.

Part of: https://github.com/paritytech/substrate/issues/12994,
Required for: https://github.com/paritytech/substrate/issues/13334

### DIFF
--- a/bounded-collections/CHANGELOG.md
+++ b/bounded-collections/CHANGELOG.md
@@ -4,7 +4,7 @@ The format is based on [Keep a Changelog].
 
 [Keep a Changelog]: http://keepachangelog.com/en/1.0.0/
 
-## [0.2.0] - 2023-05-05
+## [0.1.7] - 2023-05-05
 - Added `serde` feature, which can be enabled for no `std` deployments.
 
 ## [0.1.6] - 2023-04-27
--- a/bounded-collections/CHANGELOG.md
+++ b/bounded-collections/CHANGELOG.md
@@ -11,7 +11,6 @@ The format is based on [Keep a Changelog].
 - Added `Clone` and `Default` derive to the `impl_const_get!` macro and thereby all `Const*` types.
 - Fixed `Debug` impl for `impl_const_get!` and all `Const*` types to also print the value and not just the type name.
 
-
 ## [0.1.5] - 2023-02-13
 - Fixed `Hash` impl (previously it could not be used in practice, because the size bound was required to also implement `Hash`).
 
--- a/bounded-collections/CHANGELOG.md
+++ b/bounded-collections/CHANGELOG.md
@@ -4,6 +4,9 @@ The format is based on [Keep a Changelog].
 
 [Keep a Changelog]: http://keepachangelog.com/en/1.0.0/
 
+## [0.1.6] - 2023-05-05
+- Added `serde` feature, which can be enabled for no `std` deployments.
+
 ## [0.1.5] - 2023-02-13
 - Fixed `Hash` impl (previously it could not be used in practice, because the size bound was required to also implement `Hash`).
 
--- a/bounded-collections/CHANGELOG.md
+++ b/bounded-collections/CHANGELOG.md
@@ -4,7 +4,7 @@ The format is based on [Keep a Changelog].
 
 [Keep a Changelog]: http://keepachangelog.com/en/1.0.0/
 
-## [0.1.7] - 2023-05-05
+## [0.2.0] - 2023-05-05
 - Added `serde` feature, which can be enabled for no `std` deployments.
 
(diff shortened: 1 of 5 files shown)


bkchr (approved):

mrcnski:
Looks like there is already a `serde_no_std` feature we could reuse here: https://github.com/paritytech/parity-common/pull/385

Also, can you add a CI jobs with this feature for `bounded-collections`?

michalkucharczyk:
> Looks like there is already a `serde_no_std` feature we could reuse here: #385

Can you please elaborate more, how can we use it?
Do you have mean renaming `serde` to `serde_no_std`?

> Also, can you add a CI jobs with this feature for `bounded-collections`?
Good point, will do.

michalkucharczyk:
My personal feeling is that `serde` is just a good name for enabling this feature.  It is just about adding `serde` deps with proper features.

Naming was also discussed in https://github.com/paritytech/substrate/issues/12994:
`serde` will be used in across `substrate`.

However I do understand your point to have consistent naming in `parity-common`. Let me know what do you think?

mrcnski:
Oh, I see. Yeah, I didn't like `serde_no_std`, but changing it would be breaking and thought we should be consistent.

Is it possible to have them alias to each other? That way people using `serde_no_std` can keep using it, but we move to `serde` in substrate.

michalkucharczyk:
Nothing better than `serde = [ "serde_no_std"]` comes to my mind.
I don't know how to deprecate feature. 
Found this: https://github.com/rust-lang/cargo/issues/7130, https://stackoverflow.com/questions/56917787/how-to-correctly-deprecate-a-crate-feature.

bounded-collections/CHANGELOG.md
```diff
@@ -4,6 +4,9 @@ The format is based on [Keep a Changelog].
 
 [Keep a Changelog]: http://keepachangelog.com/en/1.0.0/
 
+## [0.1.6] - 2023-05-05
+- Added `serde` feature, which can be enabled for no `std` deployments.
```
mrcnski: nit: If we are following semver here, should this technically be a minor version bump? Since we are adding functionality.

mrcnski (approved):
Looks good, not blocking on the feature name. Maybe it's fine without further changes, will let you decide.

bounded-collections/CHANGELOG.md
michalkucharczyk: https://github.com/paritytech/parity-common/pull/745/commits/02f09de224499ab9b5dc73c6ef324660f9b5c097

mrcnski: Thanks, looks like we weren't very strict about semver in the past, but good to start now. :)

mrcnski: Actually, since we're still at `0.y.z` we might have been following the common scheme where `y` is reserved for breaking changes.

The [spec](https://github.com/semver/semver/blob/master/semver.md) recommends always bumping `y` on new releases when `x=0`, but probably we shouldn't change the strategy we were already following. Maybe we should revert? 🙈

michalkucharczyk: Both are fine for me. But following previous strategy makes sense. I'll revert.

bounded-collections/Cargo.toml
```diff
@@ -19,10 +19,12 @@ serde_json = "1.0.41"
 
 [features]
 default = ["std"]
+serde = [
+    "dep:serde"
+]
```
bkchr, line 24:
```suggestion
```

I just realized that my previous review I have overseen this :P The feature `serde` already exists, because of the optional dependency `serde`.

michalkucharczyk: ahh, right. unexplicit, but it is there...

bounded-collections/CHANGELOG.md
```diff
@@ -4,6 +4,9 @@ The format is based on [Keep a Changelog].
 
 [Keep a Changelog]: http://keepachangelog.com/en/1.0.0/
 
+## [0.1.7] - 2023-05-05
+- Added `serde` feature, which can be enabled for no `std` deployments.
```
ggwpez, line 8: Just to recapitulate: Feature `std` still enables `serde` (de)serialization and it is therefore not a breaking change.

ggwpez (approved):

bounded-collections/CHANGELOG.md
michalkucharczyk: Regarding semver: I am flexible here :)

ggwpez: No its all fine. just wanted to ensure that i understood it.

PR #767, opened 2023-07-12
Fixed clippy::incorrect_clone_impl_on_copy_type in fixed-hash

svyatonik:
There's this new lint: https://rust-lang.github.io/rust-clippy/master/index.html#/incorrect_clone_impl_on_copy_type which causes clippy failures in nightly builds. TLDR: if type is `Copy`, then `Clone` impl must be just `*self`


ordian (approved):
Thanks!

PR #781, opened 2023-08-30
Deserialize boundedbtreeset

ozgunozerk:
I followed `BoundedVec`'s Deserialize implementation. 

Tests are passing 🥳

### DIFF
--- a/bounded-collections/src/bounded_btree_set.rs
+++ b/bounded-collections/src/bounded_btree_set.rs
@@ -396,7 +396,7 @@ where
 	}
 }
 
-#[cfg(test)]
+#[cfg(all(test, feature = "std"))]
 mod test {
 	use super::*;
 	use crate::ConstU32;
--- a/bounded-collections/src/bounded_btree_set.rs
+++ b/bounded-collections/src/bounded_btree_set.rs
@@ -585,4 +585,37 @@ mod test {
 			set: BoundedBTreeSet<String, ConstU32<16>>,
 		}
 	}
+
+	#[test]
+	fn test_serializer() {
+		let mut c = BoundedBTreeSet::<u32, ConstU32<6>>::new();
+		c.try_insert(0).unwrap();
+		c.try_insert(1).unwrap();
+		c.try_insert(2).unwrap();
+
+		assert_eq!(serde_json::json!(&c).to_string(), r#"[0,1,2]"#);
+	}
+
+	#[test]
+	fn test_deserializer() {
+		let c: Result<BoundedBTreeSet<u32, ConstU32<6>>, serde_json::error::Error> = serde_json::from_str(r#"[0,1,2]"#);
+		assert!(c.is_ok());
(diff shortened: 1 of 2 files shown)


bounded-collections/src/bounded_btree_set.rs
```diff
@@ -29,9 +34,67 @@ use core::{borrow::Borrow, marker::PhantomData, ops::Deref};
...
+					Err(_) => return Err(A::Error::custom("can't convert to usize")),
+				};
+				if size > max {
+					Err(A::Error::custom("out of bounds"))
+				} else {
+					let mut values = BTreeSet::new();
+
+					while let Some(value) = seq.next_element()? {
+						values.insert(value);
+						if values.len() > max {
```
KiChjang, line 82: This is weird. You insert before checking for the max, i.e. you add 1 more value over the limit before you check the limit?

ozgunozerk: I also found it weird. It was like that in `BoundedVec`'s Deserialize implementation: https://github.com/paritytech/parity-common/blob/master/bounded-collections/src/bounded_vec.rs#L90-L92

I thought that was intentional and didn't put much thought tbh. 

I can fix them both if you like?

KiChjang: Sure, but there's one more additional issue: checking `values.len() > max` doesn't sound right to me either, because it too is also waiting for 1 element over the max before the condition triggers. What should really happen is values.len() >= max`.

It would be so great if you can write tests for these as well.

ozgunozerk: Since we are already discussing the implementation of Deserialize for BoundedVec as well, I had another thing which confused me:

https://github.com/paritytech/parity-common/blob/master/bounded-collections/src/bounded_vec.rs#L84-L85

So, if we are performing a check in here for the size, do we need another check in here: 
https://github.com/paritytech/parity-common/blob/master/bounded-collections/src/bounded_vec.rs#L90-L92
?

I think the second check is not necessary and the error produced by the second check is unreachable. So, I don't know if it's possible to write a test for triggering that error. It will trigger the previous error instead.

I might be missing something, so feel free to correct me

bkchr: The first check is using `size_hint()` and that could also be `None` in which case the size is `0` as you see in the code.

ozgunozerk: I've added a new commit:
- insertion is now performed after the bound check
- new tests for bound check:
    - i've copied the `deserializer_test` and made the size equal to the bound
    - for the tests, `size_hint()` does indeed return `None`, so I'm positive I'm testing the inner check
    - if you prefer not to have an additional test, the same test can be done by changing the bound to the size of the deserialized string in the `deserializer_test`

KiChjang (approved):

PR #827, opened 2024-02-05
Add more comment

rongyi:
make this single line more readable, guide reader the related method, they will understand this part quickly.

### DIFF
--- a/rlp/src/stream.rs
+++ b/rlp/src/stream.rs
@@ -186,6 +186,9 @@ impl RlpStream {
 			_ => {
 				// payload is longer than 1 byte only for lists > 55 bytes
 				// by pushing always this 1 byte we may avoid unnecessary shift of data
+				// both cases will need at least 1 byte header, so here, we blindly push 1 byte
+				// and then when we know the exactly size of data in we will update this value
+				// accordingly, the update detail can be found in insert_list_payload method
 				self.buffer.put_u8(0);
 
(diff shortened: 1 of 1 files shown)


rlp/src/stream.rs
```diff
@@ -186,6 +186,9 @@ impl RlpStream {
 			_ => {
 				// payload is longer than 1 byte only for lists > 55 bytes
 				// by pushing always this 1 byte we may avoid unnecessary shift of data
+				// both cases will need at least 1 byte header, so here, we blindly push 1 byte
+				// and then when we know the exactly size of data in we will update this value
+				// accordingly, the update detail can be found in insert_list_payload method
```
bkchr, line 191:
```suggestion
				// and then when we know the exactly size of data. The value will be updated
				// accordingly in `insert_list_payload` method
```

rlp/src/stream.rs
```diff
@@ -186,6 +186,9 @@ impl RlpStream {
 			_ => {
 				// payload is longer than 1 byte only for lists > 55 bytes
 				// by pushing always this 1 byte we may avoid unnecessary shift of data
+				// both cases will need at least 1 byte header, so here, we blindly push 1 byte
```
bkchr, line 189:
```suggestion
				// both cases will need at least 1 byte header, so we push 1 byte
```

bkchr (approved):

PR #853, opened 2024-07-28
Upgrade rocksdb to 0.22

jasl:
Although Parity has recommended `paritydb` for a long time, `polkadot-sdk` still has not started the deprecation process for `rocksdb`.

I think we should give some love to the rocksdb. This PR upgrade to its latest version would help to reduce a few outdated dependencies. for example, bitflags 1.3.0

### DIFF
--- a/kvdb-rocksdb/src/lib.rs
+++ b/kvdb-rocksdb/src/lib.rs
@@ -905,7 +905,7 @@ rocksdb.db.get.micros P50 : 2.000000 P95 : 3.000000 P99 : 4.000000 P100 : 5.0000
 		assert!(settings.contains(" block_size: 323232"));
 
 		// LRU cache (default column)
-		assert!(settings.contains("block_cache_options:\n    capacity : 8388608"));
+		assert!(settings.contains("block_cache_options:\n    capacity : 115343360"));
 		// LRU cache for non-default columns is ⅓ of memory budget (including default column)
 		let lru_size = (330 * MB) / 3;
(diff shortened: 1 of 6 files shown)


jasl:
@bkchr Could you help to take a look at this?

kvdb-rocksdb/src/lib.rs
```diff
@@ -905,7 +905,7 @@ rocksdb.db.get.micros P50 : 2.000000 P95 : 3.000000 P99 : 4.000000 P100 : 5.0000
 		assert!(settings.contains(" block_size: 323232"));
 
 		// LRU cache (default column)
-		assert!(settings.contains("block_cache_options:\n    capacity : 8388608"));
+		assert!(settings.contains("block_cache_options:\n    capacity : 115343360"));
```
jasl, line 908: I haven't checked why it changed here, but this should be OK.

bkchr: The cache size was probably changed? Can you maybe check that in the changelogs?

bkchr (approved):

kvdb-rocksdb/src/lib.rs
jasl: The change may relate to https://github.com/facebook/rocksdb/blob/main/HISTORY.md#public-api-changes-10

I can dig more if you think this is important

jasl:
Humbly ping @ggwpez @ordian

Could you help to review this PR?

ordian (approved):

ggwpez (approved):

PR #873, opened 2024-10-08
fixed-hash: remove `rustc-hex` feature

koushiro:
~~merge #872 first~~

- remove `rustc-hex` feature (`rustc-hex` is only used for `std::str::FromStr` impl, so I don't think the feature is necessary)
- use `const-hex` instead of `rustc-hex` dependency (`rustc-hex` is no longer maintained, and `const-hex` has [better performance](https://github.com/DaniPopes/const-hex?tab=readme-ov-file#performance))

BTW, I'm considering replacing both `hex` and `rustc-hex` in dependency tree with `const-hex`.

### DIFF
--- a/fixed-hash/Cargo.toml
+++ b/fixed-hash/Cargo.toml
@@ -9,26 +9,30 @@ description = "Macros to define custom fixed-size hash types"
 documentation = "https://docs.rs/fixed-hash/"
 readme = "README.md"
 edition = "2021"
-rust-version = "1.60"
+rust-version = "1.64.0"
 
 [package.metadata.docs.rs]
-features = ["quickcheck", "api-dummy"]
+features = ["arbitrary", "quickcheck", "api-dummy"]
 
 [dependencies]
-quickcheck = { version = "1", optional = true }
-rand = { version = "0.8.0", optional = true, default-features = false }
-rustc-hex = { version = "2.0.1", optional = true, default-features = false }
+const-hex = { version = "1.13", default-features = false }
+rand = { version = "0.8.0", default-features = false, optional = true }
 static_assertions = "1.0.0"
+
 arbitrary = { version = "1.0", optional = true }
+quickcheck = { version = "1", optional = true }
 
 [dev-dependencies]
-rand_xorshift = "0.3.0"
 criterion = "0.5.1"
 rand = { version = "0.8.0", default-features = false, features = ["std_rng"] }
+rand_xorshift = "0.3.0"
 
 [features]
-default = ["std", "rand", "rustc-hex"]
-std = ["rustc-hex/std", "rand?/std"]
+default = ["std", "rand"]
+std = ["const-hex/std", "rand?/std"]
+rand = ["dep:rand"]
+arbitrary = ["dep:arbitrary"]
+quickcheck = ["dep:quickcheck"]
 
 api-dummy = [] # Feature used by docs.rs to display documentation of hash types
 
--- a/primitive-types/Cargo.toml
+++ b/primitive-types/Cargo.toml
@@ -7,15 +7,15 @@ homepage = "https://github.com/paritytech/parity-common"
 repository = "https://github.com/paritytech/parity-common"
 description = "Primitive types shared by Ethereum and Substrate"
 edition = "2021"
-rust-version = "1.60.0"
+rust-version = "1.64.0"
 
 [dependencies]
-fixed-hash = { version = "0.8", path = "../fixed-hash", default-features = false }
-uint = { version = "0.10.0", path = "../uint", default-features = false }
-impl-serde = { version = "0.5.0", path = "impls/serde", default-features = false, optional = true }
-impl-codec = { version = "0.7.0", path = "impls/codec", default-features = false, optional = true }
-impl-num-traits = { version = "0.2.0", path = "impls/num-traits", default-features = false, optional = true }
-impl-rlp = { version = "0.4", path = "impls/rlp", default-features = false, optional = true }
+fixed-hash = { path = "../fixed-hash", version = "0.8", default-features = false }
+uint = { path = "../uint", version = "0.10.0", default-features = false }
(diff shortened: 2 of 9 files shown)


koushiro:
@ordian @bkchr PTAL

koushiro:
@ordian Is there anyone else who can review this PR?

primitive-types/Cargo.toml
```diff
@@ -7,15 +7,15 @@ homepage = "https://github.com/paritytech/parity-common"
...
-uint = { version = "0.10.0", path = "../uint", default-features = false }
-impl-serde = { version = "0.5.0", path = "impls/serde", default-features = false, optional = true }
-impl-codec = { version = "0.7.0", path = "impls/codec", default-features = false, optional = true }
-impl-num-traits = { version = "0.2.0", path = "impls/num-traits", default-features = false, optional = true }
-impl-rlp = { version = "0.4", path = "impls/rlp", default-features = false, optional = true }
+fixed-hash = { path = "../fixed-hash", version = "0.8", default-features = false }
+uint = { path = "../uint", version = "0.10.0", default-features = false }
+impl-serde = { path = "impls/serde", version = "0.5.0", default-features = false, optional = true }
+impl-codec = { path = "impls/codec", version = "0.7.0", default-features = false, optional = true }
+impl-num-traits = {  path = "impls/num-traits", version = "0.2.0",default-features = false, optional = true }
```
niklasad1, line 17:
```suggestion
impl-num-traits = { path = "impls/num-traits", version = "0.2.0",default-features = false, optional = true }
```

fixed-hash/Cargo.toml
```diff
@@ -9,26 +9,30 @@ description = "Macros to define custom fixed-size hash types"
...
 
 [package.metadata.docs.rs]
-features = ["quickcheck", "api-dummy"]
+features = ["arbitrary", "quickcheck", "api-dummy"]
 
 [dependencies]
-quickcheck = { version = "1", optional = true }
-rand = { version = "0.8.0", optional = true, default-features = false }
-rustc-hex = { version = "2.0.1", optional = true, default-features = false }
+const-hex = { version = "1.13", default-features = false }
```
niklasad1, line 18: why is `const-hex` not an optional dependency because `rustc-hex` was that before?

koushiro: if `rustc-hex` has a trait, like `rustc_hex::Trait`, and we need to impl `rustc_hex::Trait` for `H256`, then I agree that we should add `rustc-hex` as an optional feature.

niklasad1: I'm more concerned with a larger dependency tree for users that can't opt-out from  it rather than a small impl inside this crate itself...

Lots of crates in this repo are very old and I don't think had a policy when to make it optional or not. Probably that's why we haven't been consistent

Anyway, I prefer to make `const-hex` an optional dependency such that users can opt-out from it but included in the default features.

koushiro: But the most users don't learn anything from the name of the `const-hex` feature, because it is used in the implementation of `std::str::FromStr`.
so why we add an optional feature for a std trait impl, unless the impl has multiple implementations based on different features.

niklasad1: Ok, my cents just introduce a `feature hex` which brings enables `"dep:const-hex"` as an optional dependency and then do the same `uint` in this or another PR so we are consistent.

If we already are breaking things let's make it consistent...

@ordian any other ideas about this?

niklasad1: To rephrase what I mean if the feature is not part of the core functionality of the crate/type then it should be an optional feature IMO (especially if it needs extra dependencies)

I don't follow why you mean with H256 because it's re-exported from several crates such as `fixed_hash, primitive_types and ethereum_types`

For instance for ethereum-types it makes sense to have `rustc-hex` as hard dependency because it's required to be able to parse hex strings in the ethereum-world IIRC but it's not necessarily case for fixed-hash because it's depends how it used but in most use-cases it's probably useful to have.

So, if you can motivate why it's essential `From::hex_str` for these types then I'm fine to have it has hard dependency but we shouldn't use different libraries in `uint` and `fixed_hash` for those. If `const_hex` is the way to go we should use it in `uint` as well i.e, not both use `hex and const_hex`

koushiro: > I don't follow why you mean with H256 because it's re-exported from several crates such as fixed_hash, primitive_types and ethereum_types

`H256` is just a example type.

> If const_hex is the way to go we should use it in uint as well i.e, not both use hex and const_hex

see description:
> BTW, I'm considering replacing both hex and rustc-hex in dependency tree with const-hex.

koushiro: OK, let me summarize:

1. Is `core::str::FromStr` one of the core functions of `fixed-hash`? IMO, it is, so I will choose hex as its str parsing format (it will rely on a hex parsing library, whether it is `const-hex`/`hex`/`rustc-hex`, I choose `const-hex`)
2. Why do I say that the existing feature declaration (like `rustc-hex` feature) is not a good idea? IMO, features are either used for different implementations of the same function, or used to support third-party libraries. However, from the existing design of `fixed-hash`, it should not support third-party libraries through features (I think it‘s to avoid the restrictions of the orphan rule), so there are specific types declared in `primitive-types` and `ethereum-types` and a series of `impls` crates.

koushiro:
I don't want to submit PRs to a repo that doesn't give me timely feedback, so close it

fixed-hash/Cargo.toml
niklasad1: Okey, thanks for the explanation.

It's not necessarily requires to have a `FromStr` implementation to support parsing hex strings for fixed-hash but 
as you say it's quite annoying to have a feature-flag for it and makes it harder for everyone.

So, I'm onboard to make this `hex feature` required for the hash types but in the future if you want change stuff it's a much better way to open an issue to explain why something is bad and why you want to change it.

It's a more likely that we reject PRs that make breaking changes.

PR #878, opened 2024-10-24
bounded-collections: Add Const(Int|Uint) for generic const getter support

conr2d:
Refer to: paritytech/polkadot-sdk#6193

This PR introduces `ConstInt` and `ConstUint` to support generic constant integer getter.

### DIFF
--- a/bounded-collections/src/const_int.rs
+++ b/bounded-collections/src/const_int.rs
@@ -1,0 +1,156 @@
+// Copyright (C) Parity Technologies (UK) Ltd.
+//
+// Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
+// http://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+// <LICENSE-MIT or http://opensource.org/licenses/MIT>, at your
+// option. This file may not be copied, modified, or distributed
+// except according to those terms.
+
+use crate::{Get, TypedGet};
+
+trait ConstBounded<T> {
+	const MIN: T;
+	const MAX: T;
+}
+
+macro_rules! impl_const_bounded {
+	($t:ty, $type:ty) => {
+		impl ConstBounded<$type> for $t {
+			const MIN: $type = <$t>::MIN as $type;
+			const MAX: $type = <$t>::MAX as $type;
+		}
+	};
+}
+
+impl_const_bounded!(u8, u128);
+impl_const_bounded!(u16, u128);
+impl_const_bounded!(u32, u128);
+impl_const_bounded!(u64, u128);
+impl_const_bounded!(u128, u128);
+impl_const_bounded!(usize, u128);
+
+impl_const_bounded!(i8, i128);
+impl_const_bounded!(i16, i128);
+impl_const_bounded!(i32, i128);
+impl_const_bounded!(i64, i128);
+impl_const_bounded!(i128, i128);
+
+struct CheckOverflowU128<const N: u128, T: ConstBounded<u128>>(T);
+
+impl<const N: u128, T: ConstBounded<u128>> CheckOverflowU128<N, T> {
+	const ASSERTION: () = assert!(N <= T::MAX && N >= T::MIN);
+}
+
+struct CheckOverflowI128<const N: i128, T: ConstBounded<i128>>(T);
+
+impl<const N: i128, T: ConstBounded<i128>> CheckOverflowI128<N, T> {
+	const ASSERTION: () = assert!(N <= T::MAX && N >= T::MIN);
+}
+
+/// Const getter for unsigned integers.
+#[derive(Default, Clone)]
+pub struct ConstUint<const N: u128>;
+
+#[cfg(feature = "std")]
+impl<const N: u128> std::fmt::Debug for ConstUint<N> {
+	fn fmt(&self, fmt: &mut std::fmt::Formatter) -> std::fmt::Result {
+		fmt.write_str(&format!("{}<{}>", stringify!(ConstUint), N))
+	}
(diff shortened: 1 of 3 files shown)


ordian (approved):
Looks reasonable. Could you please update the changelog as well?

bounded-collections/src/lib.rs
```diff
@@ -16,13 +16,15 @@ pub extern crate alloc;
 pub mod bounded_btree_map;
 pub mod bounded_btree_set;
 pub mod bounded_vec;
+mod const_int;
```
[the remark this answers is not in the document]
conr2d: There isn't a particular reason, it's just that other ConstU* types are included in the top-level module, but since my implementation has many macros or helper types, I wrote them in a separate module solely for maintainability and readability purposes. I set them as private for that reason. If it would be better to expose them as public, I can do so.

bounded-collections/src/const_int.rs
```diff
@@ -0,0 +1,160 @@
...
+	($t:ident, $bound:ty, $target:ty) => {
+		impl<const N: $bound> Get<$target> for $t<N> {
+			fn get() -> $target {
+				let _ = <CheckOverflow<$bound>>::check::<$target, N>();
+				N as $target
+			}
+		}
+		impl<const N: $bound> Get<Option<$target>> for $t<N> {
+			fn get() -> Option<$target> {
+				let _ = <CheckOverflow<$bound>>::check::<$target, N>();
```
gui1117, line 112: This is just a runtime check no?

If I write:
```rust
pub fn foo() -> u8 {
	<ConstUint<4200> as Get<u8>>::get()
}
```
it will compile fine no?

conr2d: You're absolutely right. Initially, I wrote code that worked correctly, but in trying to refactor to make it more concise, overlooked this detail. Thanks for pointing out!

gui1117 (approved):
I never thought about asserting const expression like this.

Very nice.

gui1117:
ideally we could ensure the compilation fail with trybuild.

bounded-collections/src/const_int.rs
```diff
@@ -0,0 +1,167 @@
...
+	fn fmt(&self, fmt: &mut std::fmt::Formatter) -> std::fmt::Result {
+		fmt.write_str(&format!("{}<{}>", stringify!(ConstInt), N))
+	}
+}
+#[cfg(not(feature = "std"))]
+impl<const N: i128> core::fmt::Debug for ConstInt<N> {
+	fn fmt(&self, fmt: &mut core::fmt::Formatter) -> core::fmt::Result {
+		fmt.write_str("<wasm:stripped>")
+	}
+}
```
bkchr, line 100:
```suggestion
```

bounded-collections/src/const_int.rs
```diff
@@ -0,0 +1,167 @@
...
+
+/// Const getter for signed integers.
+#[derive(Default, Clone)]
+pub struct ConstInt<const N: i128>;
+
+#[cfg(feature = "std")]
+impl<const N: i128> std::fmt::Debug for ConstInt<N> {
+	fn fmt(&self, fmt: &mut std::fmt::Formatter) -> std::fmt::Result {
```
bkchr, line 91:
```suggestion
impl<const N: i128> core::fmt::Debug for ConstInt<N> {
	fn fmt(&self, fmt: &mut core::fmt::Formatter) -> core::fmt::Result {
```

bounded-collections/src/const_int.rs
```diff
@@ -0,0 +1,167 @@
...
+		fmt.write_str(&format!("{}<{}>", stringify!(ConstInt), N))
```
bkchr, line 92:
```suggestion
		fmt.write_str(&alloc::format!("ConstInt<{}>", N))
```

bounded-collections/src/const_int.rs
```diff
@@ -0,0 +1,167 @@
...
+	fn fmt(&self, fmt: &mut std::fmt::Formatter) -> std::fmt::Result {
+		fmt.write_str(&format!("{}<{}>", stringify!(ConstUint), N))
+	}
+}
+#[cfg(not(feature = "std"))]
+impl<const N: u128> core::fmt::Debug for ConstUint<N> {
+	fn fmt(&self, fmt: &mut core::fmt::Formatter) -> core::fmt::Result {
+		fmt.write_str("<wasm:stripped>")
+	}
+}
```
bkchr, line 76:
```suggestion
```

bounded-collections/src/const_int.rs
```diff
@@ -0,0 +1,167 @@
...
+/// # use bounded_collections::{ConstUint, Get};
+/// let _ = <ConstUint<256> as Get<u8>>::get();
+/// ```
+#[derive(Default, Clone)]
+pub struct ConstUint<const N: u128>;
+
+#[cfg(feature = "std")]
+impl<const N: u128> std::fmt::Debug for ConstUint<N> {
+	fn fmt(&self, fmt: &mut std::fmt::Formatter) -> std::fmt::Result {
+		fmt.write_str(&format!("{}<{}>", stringify!(ConstUint), N))
```
bkchr, line 68:
```suggestion
impl<const N: u128> core::fmt::Debug for ConstUint<N> {
	fn fmt(&self, fmt: &mut core::fmt::Formatter) -> core::fmt::Result {
		fmt.write_str(&format!("ConstUint<{}>", N))
```

bkchr (approved):
Nice work! Did not know about this "late const assert evaluation"

PR #887, opened 2024-12-08
Add `is_full` API to all bounded collections

shawntabrizi:
This PR adds an `is_full` API to all the bounded collections.

This was only on `BoundedVec`, but is plenty useful for all the other types too.

### DIFF
--- a/bounded-collections/src/weak_bounded_vec.rs
+++ b/bounded-collections/src/weak_bounded_vec.rs
@@ -225,6 +225,11 @@ impl<T, S: Get<u32>> WeakBoundedVec<T, S> {
 			Err(())
 		}
 	}
+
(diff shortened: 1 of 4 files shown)


bounded-collections/src/weak_bounded_vec.rs
```diff
@@ -225,6 +225,11 @@ impl<T, S: Get<u32>> WeakBoundedVec<T, S> {
 			Err(())
 		}
 	}
+
+	/// Returns true of this collection is full.
```
bkchr, line 229:
```suggestion
	/// Returns true if this collection is full.
```

bkchr (approved):

gui1117 (approved):

PR #888, opened 2024-12-19
Add `new()` function for `WeakBoundedVec`

re-gius:
Currently, to create an empty `WeakBoundedVec`, you need to do something like `WeakBoundedVec::force_from(vec![], None)`. Having a `new()` function would make it easier.


bkchr:
There is `Default` implemented. You can just use this.

PR #903, opened 2025-02-14
Upgrade rocksdb to 0.23.0

jasl:
RocksDB 0.23.0 was released on Dec 24, 2024.

Polkadot is still using 0.21.0, released on May 9, 2023. since it is still the default DB, I propose updating it to the new version. However, I don't know if we can gain performance improvement here.

I don't see it has a breaking change: https://github.com/rust-rocksdb/rust-rocksdb/releases/tag/v0.23.0

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,6 +18,8 @@ jobs:
           toolchain: stable
           override: true
 
+      - run: sudo apt-get install libclang-dev
+
       - name: Rust Cache
         uses: Swatinem/rust-cache@82a92a6e8fbeee089604da2575dc567ae9ddeaab # v2.7.5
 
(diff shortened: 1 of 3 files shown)


ordian (approved):

niklasad1 (approved):

bkchr (approved):

jasl:
> @jasl can you fix CI? looks like we will need to install some extra dependencies.

No problem, I will do later

jasl:
> @jasl can you fix CI? looks like we will need to install some extra dependencies.

CI should fixed.

FYI, I update `rust-version` to "1.71.1" to match [rust-rocksdb rust-version](https://github.com/rust-rocksdb/rust-rocksdb/blob/master/Cargo.toml#L6)

jasl:
Is it worth it to release a new version for kvdb-rocksdb?

PR #906, opened 2025-03-14
Implement `DecodeWithMemTracking` for `BoundedBTreeMap`

seadanda:
Implement `DecodeWithMemTracking` for `BoundedBTreeMap` and release bounded collections as `0.2.4`.

### DIFF
--- a/bounded-collections/CHANGELOG.md
+++ b/bounded-collections/CHANGELOG.md
@@ -4,7 +4,7 @@ The format is based on [Keep a Changelog].
 
 [Keep a Changelog]: http://keepachangelog.com/en/1.0.0/
 
-## [0.2.4] - 2025-03-14
+## [0.2.4] - 2025-03-20
 - Implement DecodeWithMemTracking for BoundedBTreeMap [#906](https://github.com/paritytech/parity-common/pull/906)
 
 ## [0.2.3] - 2025-02-11
--- a/bounded-collections/CHANGELOG.md
+++ b/bounded-collections/CHANGELOG.md
@@ -4,6 +4,9 @@ The format is based on [Keep a Changelog].
 
 [Keep a Changelog]: http://keepachangelog.com/en/1.0.0/
 
+## [0.2.4] - 2025-03-14
+- Implement DecodeWithMemTracking for BoundedBTreeMap [#906](https://github.com/paritytech/parity-common/pull/906)
+
 ## [0.2.3] - 2025-02-11
 - Implement DecodeWithMemTracking for some structs [#897](https://github.com/paritytech/parity-common/pull/897)
 
--- a/bounded-collections/Cargo.toml
+++ b/bounded-collections/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "bounded-collections"
-version = "0.2.3"
+version = "0.2.4"
 description = "Bounded types and their supporting traits"
 readme = "README.md"
 rust-version = "1.79.0"
--- a/bounded-collections/src/bounded_btree_map.rs
+++ b/bounded-collections/src/bounded_btree_map.rs
@@ -99,6 +99,39 @@ where
 	}
 }
 
+// Struct which allows prepending the compact after reading from an input.
+struct PrependCompactInput<'a, I> {
+	compact: Compact<u32>,
+	read: u32,
+	inner: &'a mut I,
+}
+
+impl<'a, I: codec::Input> codec::Input for PrependCompactInput<'a, I> {
+	fn remaining_len(&mut self) -> Result<Option<usize>, codec::Error> {
+		// Add the remaining length of the compact, if any to the inner.
+		let remaining_compact = self.compact.encode().len().saturating_sub(self.read as usize);
+		Ok(self.inner.remaining_len()?.map(|len| len + remaining_compact))
+	}
+
+	fn read(&mut self, into: &mut [u8]) -> Result<(), codec::Error> {
+		if into.is_empty() {
+			return Ok(());
+		}
+
+		let compact_bytes = self.compact.encode();
+		let remaining_compact = compact_bytes.len().saturating_sub(self.read as usize);
+
+		if remaining_compact == 0 {
+			return self.inner.read(into);
+		}
+
+		let to_read = into.len().min(remaining_compact);
+		into[..to_read].copy_from_slice(&compact_bytes[self.read as usize..][..to_read]);
+		self.read += to_read as u32;
+		Ok(())
+	}
+}
+
 impl<K, V, S> Decode for BoundedBTreeMap<K, V, S>
 where
 	K: Decode + Ord,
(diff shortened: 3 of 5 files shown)


alvicsam (approved):

bounded-collections/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "bounded-collections"
-version = "0.2.3"
+version = "0.2.4"
 description = "Bounded types and their supporting traits"
 readme = "README.md"
 rust-version = "1.79.0"
```
niklasad1, line 6: bump this to 1.81?

bounded-collections/src/bounded_btree_map.rs
```diff
@@ -123,6 +123,14 @@ where
 	}
 }
 
+impl<K, V, S> DecodeWithMemTracking for BoundedBTreeMap<K, V, S>
```
serban300, line 126: I thought that `BoundedBtreeMap` is just a wrapper over `BTreeMap`, but I see it has its own implementation for decode. So we should do the same as for in the decoding logic: `BtreeMap`: https://github.com/paritytech/parity-scale-codec/blob/8d65d28e778d914988849b786c17216070ecd9a1/src/codec.rs#L1273

bounded-collections/Cargo.toml
seadanda: I went through the workspace members and found the following rust versions to be the minimum:
```
members = [
	"fixed-hash", # 1.64
	"keccak-hash", # 1.79
	"kvdb", # 1.64
	"kvdb-memorydb", # 1.64
	"kvdb-rocksdb", # 1.71.1
	"kvdb-shared-tests", # 1.64
	"parity-bytes", # 1.64
	"rlp", # 1.64
	"rlp-derive", # 1.64
	"uint", # 1.64
	"primitive-types", # 1.79
	"bounded-collections", # 1.79
	"ethereum-types", # 1.79
	"ethbloom", # 1.64
]
```
1.64 is the minimum for the whole workspace since before this the cargo manifest requires author to be a different format.
weirdly for the individual packages I don't see the requirement for 1.81, so I wonder if there are some unused deps at the workspace level

niklasad1:
> CI should probably be using --locked, as it passed on master a month ago but fails now due to some minimum rust version requirements from dependencies.
Also should bump the rust-version of a lot of packages, since a quick look shows that some packages are at 1.56 which is unlikely to actually work. Probably best to lift it to the workspace level and set it to 1.81 for all

`--locked` doesn't make a difference right now because `Cargo.lock` is not committed to this repo because it only contains library crates. We could maybe commit Cargo.lock just to avoid such scenarios because it's possible to workaround this as a user by doing `cargo update -p zerofrom --precise <VERSION>`

bounded-collections/Cargo.toml
ggwpez: You can use [cargo msrv](https://github.com/foresterre/cargo-msrv) just to check that it works with that.  
We should also have this in CI actually.

seadanda:
Ah ok that makes sense, haven't worked in this repo before. Thanks for the context

seadanda:
WIP until https://github.com/paritytech/parity-scale-codec/pull/721 is released

bounded-collections/src/bounded_btree_map.rs
```diff
@@ -113,6 +113,7 @@ where
 			return Err("BoundedBTreeMap exceeds its limit".into());
 		}
 		input.descend_ref()?;
+		input.on_before_alloc_mem(codec::mem_size_of_btree::<(K, V)>(len))?;
 		let inner = Result::from_iter((0..len).map(|_| Decode::decode(input)))?;
 		input.ascend_ref();
```
bkchr, line 118: Should be replaced by

```
struct PrependCompactInput { compact: Compact<u32>, read: u32, inner: I }

/// impl Input that first prepends the compact and then uses the inner

/// After that you use it

let inner = BTreeMap::decode(&mut PrependCompactInput { compact, read: 0, inner: input })?
```

bkchr (changes requested):

bounded-collections/src/bounded_btree_map.rs
seadanda: Nice idea, implemented as you've described it in [1a73132](https://github.com/paritytech/parity-common/pull/906/commits/1a7313228a8b930a28ac85e440efee22a69a21eb)

Depending on how hard we want to go we could maybe avoid some allocations since `Compact<u32>` has a fixed upper limit on its size with something like:

```
struct PrependCompactInput<'a, I> {
    compact_bytes: [u8; 5],
    compact_len: u8,
    read: u32,
    inner: &'a mut I,
}
```
Then you just encode the compact once instead of on each read. When we construct the struct we dump the compact into an array and hardcode its length, so we know only the first `compact_len` bytes are valid

bounded-collections/src/bounded_btree_map.rs
```diff
@@ -99,22 +99,53 @@ where
...
+
+		let compact_bytes = self.compact.encode();
+		let remaining_compact = compact_bytes.len().saturating_sub(self.read as usize);
+
+		if remaining_compact == 0 {
+			return self.inner.read(into);
+		}
+
+		let to_read = into.len().min(remaining_compact);
+		into[..to_read].copy_from_slice(&compact_bytes[self.read as usize..][..to_read]);
```
bkchr, line 129:
```suggestion
		into[..to_read].copy_from_slice(&self.encoded_len[self.read..][..to_read]);
```

bounded-collections/src/bounded_btree_map.rs
```diff
@@ -99,22 +99,53 @@ where
...
+		self.read += to_read as u32;
```
bkchr, line 130:
```suggestion
		self.read += to_read ;
```

bounded-collections/src/bounded_btree_map.rs
```diff
@@ -99,22 +99,53 @@ where
...
+		}
+
+		let compact_bytes = self.compact.encode();
+		let remaining_compact = compact_bytes.len().saturating_sub(self.read as usize);
+
+		if remaining_compact == 0 {
+			return self.inner.read(into);
+		}
+
+		let to_read = into.len().min(remaining_compact);
```
bkchr, line 128: You also need to have loop, because `into.len()` is mayber bigger than `remaining_compact`. The `read` function requires that all the requested bytes are filled.

seadanda: Just got around to fixing this in [908af22](https://github.com/paritytech/parity-common/pull/906/commits/908af22b482f19a1dc59ee29b94d6fa42675da07) and added tests. Should be good to go now

bounded-collections/CHANGELOG.md
```diff
@@ -4,6 +4,9 @@ The format is based on [Keep a Changelog].
 
 [Keep a Changelog]: http://keepachangelog.com/en/1.0.0/
 
+## [0.2.4] - 2025-03-14
```
bkchr, line 7:
```suggestion
## [0.2.4] - 2025-03-20
```

bkchr (approved):
Ty Ser!

bkontur:
Hey guy, when we will bring this to the master?
https://github.com/paritytech/polkadot-sdk/blob/master/Cargo.toml#L655
```
bounded-collections = { version = "0.2.3", default-features = false }
```
0.2.4 is already released: https://crates.io/crates/bounded-collections/0.2.4

Are we waiting for anything?

seadanda:
No not waiting for anything - the work that I needed this for is still on a dev branch until we get the migration working and tested for Westend.

It can be bumped on master without any issues

PR #914, opened 2025-05-21
Support JAM codec for bounded collections

davxy:
Previously, `bounded-collections` was tightly coupled to `scale-codec`, which posed limitations as our usage evolved.

For now, in **PolkaJam**, we're relying on a temporary fork: [`bounded-collections-next`](https://crates.io/crates/bounded-collections-next) with these changes applied. However, our long-term goal is to have `jam-codec` supported upstream.

---

This PR introduces the following changes:

- Makes the choice of codec optional (though we might retain `scale-codec` as the default for backward compatibility).
- Uses feature flags and macros to conditionally enable codec-specific code, as `jam-codec` and `scale-codec` share an identical API (different in their implementations of integers compact encoding).

### DIFF
--- a/bounded-collections/Cargo.toml
+++ b/bounded-collections/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "bounded-collections"
-version = "0.2.4"
+version = "0.3.0"
 description = "Bounded types and their supporting traits"
 readme = "README.md"
 rust-version = "1.79.0"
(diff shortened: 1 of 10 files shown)


bounded-collections/Cargo.toml
```diff
@@ -12,10 +12,11 @@ repository.workspace = true
 
 [dependencies]
 serde = { workspace = true, features = ["alloc", "derive"], optional = true }
-codec = { workspace = true, features = ["max-encoded-len"] }
-scale-info = { workspace = true, features = ["derive"] }
+scale-codec = { workspace = true, default-features = false, features = ["max-encoded-len"], optional = true }
+scale-info = { workspace = true, features = ["derive"], optional = true }
 log = { workspace = true }
 schemars = { workspace = true, optional = true }
+jam-codec = { version = "0.1.0", default-features = false, features = ["derive","max-encoded-len"], optional = true }
```
bkchr, line 19: Please put this to the workspace, as all the other dependencies.

bkchr (approved):
Besides the dependency that should be moved, it looks good.

arkpar (approved):

PR #915, opened 2025-05-28
Implement `DecodeWithMemTracking` for `BoundedBTreeSet`

dimartiro:
# Description

Implement `DecodeWithMemTracking` for `BoundedBTreeMap`

### DIFF
--- a/bounded-collections/src/bounded_btree_set.rs
+++ b/bounded-collections/src/bounded_btree_set.rs
@@ -359,7 +359,9 @@ where
 macro_rules! codec_impl {
 	($codec:ident) => {
 		use super::*;
-		use $codec::{Compact, Decode, DecodeLength, Encode, EncodeLike, Error, Input, MaxEncodedLen};
+		use $codec::{
+			Compact, Decode, DecodeLength, DecodeWithMemTracking, Encode, EncodeLike, Error, Input, MaxEncodedLen,
+		};
 		impl<T, S> Decode for BoundedBTreeSet<T, S>
 		where
 			T: Decode + Ord,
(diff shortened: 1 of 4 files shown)


bkchr (approved):

serban300:
> Could you please point me to where `on_before_alloc_mem` is called in that PR?

It's not specifically called in that PR. It's called in `BTreeMap::decode()` which is used in the PR: https://github.com/paritytech/parity-scale-codec/blob/5c51d86fc3d105e4846f3c0ddb22d26d6d11eea4/src/codec.rs#L1276

And for `BtreeSet` it's the same

bkchr (changes requested):
Please add some test

serban300:
Sorry, I was a bit busy last week. I will try to take a look these days

bounded-collections/src/bounded_btree_set.rs
```diff
@@ -359,22 +359,28 @@ where
...
 			T: Decode + Ord,
 			S: Get<u32>,
 		{
 			fn decode<I: Input>(input: &mut I) -> Result<Self, Error> {
-				// Same as the underlying implementation for `Decode` on `BTreeSet`, except we fail early if
-				// the len is too big.
-				let len: u32 = <Compact<u32>>::decode(input)?.into();
-				if len > S::get() {
+				// Fail early if the len is too big. This is a compact u32 which we will later put back.
+				let compact = <Compact<u32>>::decode(input)?;
```
serban300, line 374: Nit: I would rename this to `len` since it holds the len of the collection, even though it's represented as a `Compact`. Also the same for `BTreeMap`.

serban300 (commented):
LGTM ! Just left a nit.

bkchr (approved):

bounded-collections/src/codec_utils.rs
```diff
@@ -0,0 +1,153 @@
...
+
+					// And we can't read more.
+					assert!(input.read(&mut buf).is_err());
+				}
+			}
+		};
+	}
+
+	// Generate tests for each available codec
+    #[cfg(feature = "scale-codec")]
```
bkchr, line 149:
```suggestion
	#[cfg(feature = "scale-codec")]
```

serban300:
I can do that, but probably tomorrow

PR #934, opened 2025-08-31
Update `impl_const_get` macro to allow getting larger types

shawntabrizi:
This update implements smaller types to implement `Get` for larger types.

For example, a `ConstU8` will be able to provide `Get<u32>`.

This can help if you are making a `BoundedVec`, where now you can use smaller types like `ConstU8` as the bound.

### DIFF
--- a/bounded-collections/src/lib.rs
+++ b/bounded-collections/src/lib.rs
@@ -93,7 +93,7 @@ impl<T: Default> Get<T> for GetDefault {
 }
 
 macro_rules! impl_const_get {
-	($name:ident, $t:ty) => {
+	($name:ident, $t:ty, get_into: [$($larger:ty),*]) => {
 		/// Const getter for a basic type.
 		#[derive(Default, Clone)]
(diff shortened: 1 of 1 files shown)


bounded-collections/src/lib.rs
```diff
@@ -93,7 +93,7 @@ impl<T: Default> Get<T> for GetDefault {
 }
 
 macro_rules! impl_const_get {
-	($name:ident, $t:ty) => {
+	($name:ident, $t:ty, get_into: [$($larger:ty),*]) => {
```
bkchr, line 96: You can simplify the implementation and let it do the compiler: https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=a77851b761e90d2246deb9e53395c734

gui1117 (approved):
it may break some type inference. But the code is good.

bkchr:
Yeah the next release should be a breaking release.

Issue #932, opened 2025-08-15
Publish new version


StackOverflowExcept1on:
https://github.com/rust-rocksdb/rust-rocksdb/issues/995
I can't compile RocksDB on Ubuntu 25.10

serban300:
https://github.com/paritytech/parity-common/pull/935

serban300:
Published `kvdb-rocksdb v0.20.0`: https://crates.io/crates/kvdb-rocksdb/0.20.0

PR #936, opened 2025-09-03
Fix `kvdb-rocksdb` publish

serban300:
Related to https://github.com/paritytech/parity-common/issues/932

Fixes the following publish error:

```
error: all dependencies must have a version specified when publishing.
dependency `kvdb` does not specify a version
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -74,7 +74,7 @@ impl-codec = { path = "./primitive-types/impls/codec", default-features = false 
 impl-num-traits = { path = "./primitive-types/impls/num-traits", default-features = false }
 impl-rlp = { path = "./primitive-types/impls/rlp", default-features = false }
 impl-serde = { path = "./primitive-types/impls/serde", default-features = false }
-kvdb = { path = "./kvdb" }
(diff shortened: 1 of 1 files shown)


Cargo.toml
```diff
@@ -74,7 +74,7 @@ impl-codec = { path = "./primitive-types/impls/codec", default-features = false
 impl-num-traits = { path = "./primitive-types/impls/num-traits", default-features = false }
 impl-rlp = { path = "./primitive-types/impls/rlp", default-features = false }
 impl-serde = { path = "./primitive-types/impls/serde", default-features = false }
-kvdb = { path = "./kvdb" }
+kvdb = { path = "./kvdb", version = "0.13" }
```
bkontur, line 77: Why others do not need version? E.g. `ethereum-types = { path = "./ethereum-types" }`

bkontur (approved):

Cargo.toml
serban300: Probably we didn't need to publish a package that depends on it lately. I think it's best to fix each dependency when we hit the issue. If I try to fix everything in this PR, 'm afraid I might miss something

PR #939, opened 2025-09-18
primitive-types: Use correct scale info type for `U256` and `U128`

bkchr:
Fixes: https://github.com/paritytech/polkadot-sdk/issues/9773


serban300 (approved):

bkchr:
> about `pub struct U512(8);`, it stays `[u64; 8]`, because there is no corespondent variant in

Yes exactly.

PR #975, opened 2026-04-10
bounded_vec macro + hash in no-std

gui1117:
This PR does:
- add `Hash` implementation on some collection even in no-std environment
- make `bounded_vec!` and `bounded_btree_map!` macro available in no-std environment
- run clippy on the crate.


gui1117:
I'm realizing that bounded_vec macro may be std only because it simply unwraps, which is bad.

And there is not way to get the bound at compile time.

PR #979, opened 2026-07-08
kvdb-rocksdb: optional per-column storage path

x3c41a:
Adds `ColumnConfig.path` so a single column family's SST files can be placed on a separate directory, letting a large cold column live on a cheaper volume while the rest of the DB stays on the main disk.

When `path` is set, `column_config` gives that column a single `cf_paths` entry with a max target size, which pins the column to that directory (RocksDB does not spill elsewhere, and `cf_paths` overrides `db_paths` for that column). Empty by default, so existing databases and single-volume setups are unaffected. `column_config` now returns `io::Result` to surface an invalid path.

Motivation: Polkadot Bulletin Chain wants to keep its bulk indexed-transaction column on cheap SSD/HDD while consensus state stays on NVMe.

Note: `cf_paths` affects only where new SST files are written; relocating an already-populated column needs a compaction or resync.

Depends on rust-rocksdb exposing `Options::set_cf_paths` (rust-rocksdb/rust-rocksdb#1094); compiles and runs against that change locally.

Draft for discussion.


dr333ws:
Heads up on CI: the compile failure (`no method named set_cf_paths`) is expected for now. `Options::set_cf_paths` was merged upstream in rust-rocksdb#1092 but is not in a published `rocksdb` release yet (latest crate is 0.24.0, which predates the merge). This PR will build once rust-rocksdb cuts a release including #1092 and the `rocksdb` dependency here is bumped to it. Keeping this as a draft until then. (Rustfmt is now fixed.)

PR #980, opened 2026-07-18
Fix trailing semicolon in expression macro bodies

nazar-pc:
This is a trivial change, which addresses https://github.com/rust-lang/rust/issues/79813.

It is going to become a hard error soon: https://github.com/rust-lang/rust/pull/159218.

The only reason is wasn't caught earlier is due to https://github.com/rust-lang/rust/pull/159222, which is how I noticed it after Nightly upgrade.

I would also appreciate a patch version released shortly after merging this.


nazar-pc:
Ping here, please take a look

nazar-pc:
@bkchr maybe?

nazar-pc:
Patch release, please :pray:

bkchr:
@nazar-pc sorry, for the delay. 0.10.1 is released!

